PR #29, opened 2025-05-19
Session error debuging

rdubigny:


rdubigny:
[Investigation : Erreur Y190004 - sessionInvalidSession](https://trello.com/c/ToAJrHvb/635-investigation-erreur-y190004-sessioninvalidsession)

back/apps/core-fca/src/decorators/user-session.decorator.ts
```diff
@@ -42,6 +42,11 @@ export const UserSessionDecoratorFactory = async (
     const mandatoryPropertiesErrors = await validate(object as object);
 
     if (mandatoryPropertiesErrors.length) {
+      console.log({
```
gdarquie, line 45: question: Why aren't you using the logger from the logger lib for logging?

rdubigny: Because it's too much code for the same result. I would have used the logger if the code were meant to last more than three weeks.

PR #43, opened 2025-05-28
refactor: move error message generation in a dedicated exception filter

rdubigny:


Morendil:
What I gather this PR is trying to do is make the contactHref portion of CoreFcaInvalidIdentityException more readable (it's a mess at the moment).

I've been spending some time in the exception filters and based on that experience I'd recommend against adding a new filter. They have lots of complexity…

What I see is that:
- the error page displays a specific section if the exception has the field `displayContact` set to `true`.
- this specific section allows the user to report feedback, it uses three pieces of information:
  - a specific call to action message, just for this section
  - a contact email (who to report the error to), which can be either PC support, the service provider, or the identity provider
  - an email body (optional, only used in the invalid identity case)

There are only four exceptions with `displayContact = true` (other than CoreFcaBaseException):
- CoreFcaAgentNotFromPublicServiceException
- CoreFcaInvalidIdentityException
- CoreFcaAgentNoIdpException
- CoreFcaUnauthorizedEmailException

In addition, CoreFcaBaseException has `displayContact = true` and this is then set back to `false` by two additional exceptions:
- CoreFcaAgentAccountBlockedException
- CoreFcaAgentIdpDisabledException

All of the exceptions involved are Core exceptions, and their intended context is the error web page (as opposed to a JSON result) so the only filter where they make sense is FcWebHtmlExceptionFilter.

We only use the service provider contact email in `CoreFcaUnauthorizedEmailException`.

(Details: It is passed by constructor from "temporary code for resolving Uniforces issue" in `CoreFcaService.validateEmailForSp`, which is called from the redirect to idP client route. The error seems intended to nudge (if it can be called that) users by reminding them to use a professional email address rather than a personal one.)

We only use the identity provider contact in `CoreFcaInvalidIdentityException`.

(Details: it is passed by constructor from `IdentitySanitizer.getValidatedIdentityFromIdp` in the context of the OIDC Callback route.)

It feels quite messy to be getting this information into the exception object, where it does not really belong. The information that needs to be in the exception is the reason for the problem: what is the email address that we think is inappropriate, or what is wrong with the identity returned by the identity provider.

So my plan would be:
- do everything in FcWebHtmlExceptionFilter (get IdP and SP contact info, format the messages…)
- use only a single "call to action" message, there's no real need for five different ones
- make a special "tag" interface for the exceptions that need displayContact behaviour
- make the exception classes simpler with only the information they need
- create an email body template (.ejs) to make the "invalid identity" message easier to read and modify

How does that sound ?

PR #44, opened 2025-05-28
[PCF-650] Add pm2 on Dockerfile to debug log rotate

gdarquie:
In addition to adding PM2 to the Dockerfile, we compared it with previous fc-docker Dockerfiles to incorporate any commands or configurations that seemed useful.


douglasduteil (approved):
LGTM

PR #47, opened 2025-05-30
Refactor idp identity tranformations

rdubigny:


back/apps/core-fca/src/dto/identity-for-sp.dto.ts
```diff
@@ -1,91 +1,27 @@
...
-  @MaxLength(256)
-  @IsOptional()
-  @Expose()
-  readonly siren?: string;
-
-  @IsSiret()
-  @Expose()
-  readonly siret: string;
+  @IsPhoneNumberSimpleValidator({ groups: ['phone_number'] })
+  phone_number?: string;
```
gdarquie, line 15: For the phone number transformation, maybe we could do all operations inside the dto?

```
  @Transform(({ value }) => {
    // Use your custom phone number validation logic here
    const isValidPhoneNumber = IsPhoneNumberFca(value);
    return isValidPhoneNumber ? value : undefined;
  })
  @IsPhoneNumberFca()
  @IsOptional()
  @Expose()
  readonly phone_number?: string;

```

back/apps/core-fca/src/controllers/oidc-client.controller.ts
```diff
@@ -319,42 +315,26 @@ export class OidcClientController {
       idpId,
     };
 
-    const identity =
-      await this.oidcClient.getUserInfosFromProvider<IdentityFromIdpDto>(
-        userInfoParams,
-        req,
-      );
+    const plainIdpIdentity = await this.oidcClient.getUserinfo(userInfoParams);
```
gdarquie, line 318: thought: if we could remove all the wrapping of getUserInfo() here, it would be neat, this controller calls oidcClient.getUserinfo

```js
  async getUserinfo({ accessToken, idpId }: UserInfosParams): Promise<any> {
    // OIDC: call idp's /userinfo endpoint
    try {
      return await this.utils.getUserInfo(accessToken, idpId);
    } catch (error) {
      this.logger.debug(error.stack);
      throw new OidcClientUserinfosFailedException();
    }
  }
```
It is a wrapping of await this.utils.getUserInfo(accessToken, idpId)

And the function getUserInfo is the following: 

```js
  async getUserInfo<T>(accessToken: string, idpId: string): Promise<T> {
    const client = await this.issuer.getClient(idpId);

    const userInfo = await client.userinfo<T>(accessToken);

    return userInfo;
  }
```

Maybe we could put that in a separated commit, but I think it could help us to stop to use the utils.service.

back/apps/core-fca/src/dto/identity-for-sp.dto.ts
```diff
@@ -1,91 +1,27 @@
...
-import {
-  IsAscii,
-  IsBoolean,
-  IsEmail,
-  IsOptional,
-  IsString,
-  MaxLength,
-  MinLength,
-} from 'class-validator';
+import { IsObject, IsString } from 'class-validator';
```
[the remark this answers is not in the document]
gdarquie: I've created a PR with this branch as target as a demo: https://github.com/proconnect-gouv/federation/pull/64/files

rdubigny:
will be merged with https://github.com/proconnect-gouv/federation/pull/65

PR #48, opened 2025-05-30
[PCF-674]📦 Remove useless npm packages "luxon" and "nodemailer"

gdarquie:
When removing the unused packages, new errors were generated in the CI.
While debugging, I realized that the CI cache system was always being triggered, followed by npm install.
For now, I've removed the cache system to keep the CI as clean and simple as possible.


gdarquie:
[Supprimer les packages inutiles ajoutés](https://trello.com/c/p6Vu461W/674-supprimer-les-packages-inutiles-ajout%C3%A9s)

PR #49, opened 2025-05-30
[PC-2176] Remove sp-mock from nest monorepo

gdarquie:


gdarquie:
Someday, a courageous person will save us from this infamy: mocks are not part of the Nest monorepo, yet they still live inside it.

PR #54, opened 2025-06-02
feat: add proconnect-identity acr levels

rdubigny:


rebeccadumazert (approved):
It works well even if you don't use "https://**www**.proconnect.gouv.fr/assurance/blabla" ?

PR #92, opened 2025-06-12
add possibility to have none algorithm in mock FI

BenoitSerrano:


gdarquie:
question: Does this mean there's nothing that needs to be done to make ProConnect accept user info without a signed response? The only changes here concern the Docker stack and the mocks.

This PR proves that Federation can have some identity providers that returns user info without a signed response but nothing has been changed in its configuration, isn't it?

gdarquie:
[Désactiver la signature obligatoire sur le user info pour les FI](https://trello.com/c/938qtMym/105-d%C3%A9sactiver-la-signature-obligatoire-sur-le-user-info-pour-les-fi)

BenoitSerrano:
> question2: we have the same thing to do for the service provider (i.e: add the possibility to have no sign response), isn't it?

Nope, as far as I know it's not something that the SP have asked of us

gdarquie (approved):

PR #97, opened 2025-06-16
Remove exception translation mappings

Morendil:
The lines in `back/libs/oidc-provider/src/exceptions/runtime/fr-FR.i18n.ts` are all identical and add no value.


Morendil:
Superseded by #99

PR #98, opened 2025-06-16
[PCF-698]💬 Nginx logs are also displayed on standard output

gdarquie:


gdarquie:
[Faire sortir les logs nginx sur la sortie standard en dev](https://trello.com/c/yyJueqIT/698-faire-sortir-les-logs-nginx-sur-la-sortie-standard-en-dev)

PR #99, opened 2025-06-16
[PCF-702]🔥 Remove oidc generated runtimes errors

gdarquie:
What's going on in this PR ? Quite a lot !
- remove all of the « translation layer exceptions » that wrapped OIDC errors
- **the error code for OIDC exceptions is now the exception class and no longer an Yxxxxxx code**
- remove the OIDC errors documentation page (which didn't add lots of value)
- remove generation of that page in `export-exceptions`
- remove the translation layer for OIDC exception error messages (which added no value as all messages were the default)
- refactor the BaseExceptiond and OidcProviderNoWrapperException to give them some behaviour (previously in functions in the Filter classes)
- adjust E2E and unit tests


quality/fca/cypress/support/api/steps/api-response-steps.ts
```diff
@@ -116,14 +116,17 @@ Then('le corps de la réponse contient un JWT', function () {
...
+Then(
+  'le corps de la réponse contient une erreur avec les champs error et error_description',
+  function () {
+    cy.get('@apiResponse')
+      .its('body')
+      .then(async (body) => {
+        const errors = await validateDto(body, ApiErrorDto, {
+          forbidNonWhitelisted: true,
+          whitelist: true,
+        });
```
rdubigny, line 128: suggestion: can we avoid usage of validateDto and use the underlying functions ?

```suggestion
        const error = plainToInstance(ApiErrorDto, body);
        const validationErrors = await validate(error, {
          forbidNonWhitelisted: true,
          whitelist: true,
        });
```

quality/fca/cypress/support/api/steps/api-response-steps.ts
```diff
@@ -116,14 +116,17 @@ Then('le corps de la réponse contient un JWT', function () {
...
+        expect(errors, JSON.stringify(errors)).to.have.length(0);
```
rdubigny, line 129: Suggestion: actually the use of class validator seems over-engineered to me.

I suggest you just check that erro and error description are defined and delete ApiErrorDto.

Morendil: OK, ApiErrorDto and this call to validateDto are no more !

Morendil:
(Closing and reopening to see if it gets the CI checks unstuck from "Expected - Waiting for status to be reported"…)

back/libs/exceptions/src/exceptions/base.exception.ts
```diff
@@ -16,6 +18,31 @@ export class BaseException extends Error {
   public statusCode?: number;
   public status?: number;
 
+  public getMessage(): string {
```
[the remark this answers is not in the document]
Morendil: I've done both:
- next to last commit removes the i18n lib entirely
- last commit moves behavior out of BaseException and OidcProviderNoWrapperException

…I personally think the result of the last commit is less readable than the other way, it leads to a bigger mess of conditionals.

PR #140, opened 2025-06-25
Catch errors in oidc-callback endpoint

Morendil:
The main point of this PR is to improve the error page display of some errors, which was previously misleading. Errors returned via the identity provider through the callback endpoint would be displayed instead as a "missing code" (Y020025) error, when this was not the actual problem.

With this PR we
- add a facility in the mock identity provider to simulate the behaviour where the IdP terminates the login flow abnormally by calling the callback URI with `error` and `error_description` parameters
- add a step to the E2E machinery which takes advantage of this, and use it to test that we:
- keep simplifying the handling of exceptions, notably by removing getParams() in filters and their specs
- make the unit tests more discriminating of the exception handling logic, removing a few mocks along the way
- remove checkCode from OidcClientUtilsService, and let oidc-client raise an exception
- upon catching this exception **display the actual error in the error page**, logged (and other usual behaviours for an exception)


back/apps/core-fca/src/controllers/oidc-client.controller.ts
```diff
@@ -275,6 +276,21 @@ export class OidcClientController {
     @UserSessionDecorator(GetOidcCallbackSessionDto)
     userSession: ISessionService<UserSession>,
   ) {
+    if (req.query['error']) {
+      const dict = {};
+      dict[req.query['error']] = req.query['error_description'];
```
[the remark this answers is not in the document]
jonathanperret: Thinking about it a bit more, I don't think this is susceptible to prototype pollution, but it still feels weird to have the name of an object key be controlled by the client.
Looking at the context, it seems this `dictionary` member only exists to satisfy a vestigial requirement in `error.ejs` — maybe removing that in favor of a direct `error_description` member might be saner?

back/apps/core-fca/src/controllers/oidc-client.controller.ts
```diff
@@ -275,6 +276,21 @@ export class OidcClientController {
...
+      res.status(400);
+      return res.render('error', {
+        exception: {},
+        error: {
+          code: req.query['error'],
+          id: uuid(),
```
jonathanperret, line 287: As far as I know, the point of generating unique IDs for errors is to allow support personnel to look up more details (e.g. stack traces) in a log or event tracking system when a user reports an error they encountered.

Generating a random ID that is not otherwise logged anywhere feels quite pointless unless I missed something.

back/apps/core-fca/src/controllers/oidc-client.controller.ts
```diff
@@ -275,6 +276,21 @@ export class OidcClientController {
     @UserSessionDecorator(GetOidcCallbackSessionDto)
     userSession: ISessionService<UserSession>,
   ) {
+    if (req.query['error']) {
```
jonathanperret, line 279: It might be a good idea to use Nest's `@Query` decorator instead of `req.query`.

Morendil: Done.

rdubigny (commented):
I have a major concern about this PR. I let you read my comment. I am available for discussion.

Morendil:
Recreated in #220

PR #156, opened 2025-07-01
refactor: use transformed object

rdubigny:
The previous implementation did not whitelist object property as the result of validation was droped. Also no transformation by class-transformer was possible for the same reason.

Proposal

Use the class-transformer and class-validator results as object passed to the application.

### DIFF
--- a/back/libs/data-provider-adapter-mongo/src/services/data-provider-adapter-mongo.service.ts
+++ b/back/libs/data-provider-adapter-mongo/src/services/data-provider-adapter-mongo.service.ts
@@ -1,13 +1,12 @@
 import { plainToInstance } from 'class-transformer';
 import { validate } from 'class-validator';
 import * as deepFreeze from 'deep-freeze';
-import { cloneDeep } from 'lodash';
+import { cloneDeep, filter } from 'lodash';
 import { Model } from 'mongoose';
 
 import { Injectable } from '@nestjs/common';
 import { InjectModel } from '@nestjs/mongoose';
 
-import { asyncFilter } from '@fc/common';
 import { ConfigService } from '@fc/config';
 import { CryptographyService } from '@fc/cryptography';
 import { LoggerService } from '@fc/logger';
@@ -82,7 +81,7 @@ export class DataProviderAdapterMongoService implements IDataProviderAdapter {
   }
 
   private async findAllDataProvider(): Promise<DataProviderMetadata[]> {
-    const rawResult = await this.dataProviderModel
+    const rawDataProviders = await this.dataProviderModel
       .find(
         {},
         {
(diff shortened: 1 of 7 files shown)


back/libs/data-provider-adapter-mongo/src/services/data-provider-adapter-mongo.service.ts
```diff
@@ -91,32 +90,31 @@ export class DataProviderAdapterMongoService implements IDataProviderAdapter {
...
+        if (errors.length) {
           this.logger.alert({
             msg: `Data provider "${uid}" was excluded at DTO validation`,
             validationErrors: errors,
           });
         }
 
-        return errors.length === 0;
-      },
+        return !errors.length ? dataProvider : undefined;
```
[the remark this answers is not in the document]
BenoitSerrano:
```suggestion
        return errors.length === 0 ? dataProvider : undefined;
```

back/libs/fqdn-to-idp-adapter-mongo/src/service/fqdn-to-idp-adapter-mongo.service.ts
```diff
@@ -80,48 +81,46 @@ export class FqdnToIdpAdapterMongoService
...
         );
-
-        if (errors.length > 0) {
-          this.logger.warning(
-            `fqdnToProvider with domain "${doc.fqdn}" and provider uuid "${doc.identityProvider}" was excluded from the result at DTO validation.`,
-          );
-          this.logger.err({ errors });
+        const errors = await validate(fqdnToIdentityProvider, {
+          forbidNonWhitelisted: true,
+          skipMissingProperties: false,
```
Morendil, line 106: false seems to be the default, but we may want to keep it for explicitness

back/libs/data-provider-adapter-mongo/src/services/data-provider-adapter-mongo.service.ts
```diff
@@ -91,32 +90,31 @@ export class DataProviderAdapterMongoService implements IDataProviderAdapter {
       )
       .lean();
 
-    const dataProviders = await asyncFilter<DataProviderMetadata[]>(
-      rawResult,
-      async (dataProviderRaw: any) => {
+    const dataProviders = await Promise.all(
```
jonathanperret, line 93: Here's a proposal for writing this without `Promise.all`, without `lodash.filter` and without `any`:

```suggestion
    const dataProviders = [];

    for (const rawDataProvider of rawDataProviders) {
      const dataProvider = plainToInstance(
        DataProviderAdapterMongoDTO,
        rawDataProvider,
      );
      const errors = await validate(dataProvider, {
        forbidNonWhitelisted: true,
        skipMissingProperties: false,
        whitelist: true,
      });
      
      if (errors.length) {
        const { uid } = rawDataProvider;
        this.logger.alert({
          msg: `Data provider "${uid}" was excluded at DTO validation`,
          validationErrors: errors,
        });
      } else {
        dataProviders.push(dataProvider);
      }
    }
```

Note that `Promise.all` is a red flag when applied to unconstrained collections (here, the size of `rawDataProviders` is unbounded) because it implies that an unbounded number of promises have been created, which in turn means that a possibly large number of asynchronous tasks have been launched simultaneously.

It's generally better for performance and memory to process items sequentially (as above thanks to `for` + `await`), or with controlled concurrency by using something like https://github.com/sindresorhus/p-map which allows limiting the number of promises created concurrently.

In this particular case it seems the `validate` calls will return synchronously in practice, so it's very likely that sequential processing is the best option in terms of overhead.

rdubigny: brilliant

PR #161, opened 2025-07-02
🔨 dks prune-all only remove docker contents

gdarquie:
Actually, the `dks prune-all` command requires sudo access.
I propose modifying it to only handle Docker content deletion, removing the need for elevated privileges.


gdarquie:
Closed because it was not accepted in its current form and requires further discussion. This PR was initially intended as a quick win.

We can reopen it later. One suggestion from @jonathanperret was to change the responsibility of this command to focus solely on cleaning the project — continuing to delete npm modules, but stopping short of running docker system prune.

PR #163, opened 2025-07-02
ci(github): pin to proconnect-gouv/federation-admin@pc-359

douglasduteil:


douglasduteil-admin:
latest `federation-admin` won't break federation

PR #200, opened 2025-07-10
⚡ ci(back): add yarn cache optimization to GitHub workflows

douglasduteil:
<div align=center></div>

---

Added yarn caching to improve CI performance by caching dependencies based on yarn.lock files.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -19,6 +19,8 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
+          cache: yarn
+          cache-dependency-path: ./back/yarn.lock
           node-version-file: ./back/package.json
 
       - name: Install dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-tests.yml
```diff
@@ -73,6 +77,8 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
+          cache: yarn
+          cache-dependency-path: ./back/yarn.lock
```
gdarquie, line 81: question: @douglasduteil I cannot find any reference of `cache-dependency-path` in https://github.com/actions/cache or https://docs.github.com/en/actions/reference/dependency-caching-reference.

We don't use anymore a key like

```
      - name: Cache Node.js dependencies
        uses: actions/cache@v3
        with:
          path: node_modules
          key: ${{ runner.os }}-node-${{ hashFiles('**/yarn.lock') }}
          restore-keys: |
            ${{ runner.os }}-node-
```

douglasduteil: @gdarquie 
It's because the cache is handle by https://github.com/actions/setup-node see https://github.com/actions/setup-node?tab=readme-ov-file#caching-global-packages-data

PR #217, opened 2025-07-11
🎨 Simplify app startup docker process in development

gdarquie:


gdarquie:
Divided in many small PR we can find connected to this trello ticket: https://trello.com/c/W5r7Q6Qw/710-simplification-du-lancement-des-stacks

PR #220, opened 2025-07-15
Catch IdP errors in oidc-callback endpoint

Morendil:
(Recreates #140 manually to work around problems arising from rebasing with the unsigned commits from admin.)

The main point of this PR is to improve the error page display of some errors, which was previously misleading. Errors returned via the identity provider through the callback endpoint would be displayed instead as a "missing code" (Y020025) error, when this was not the actual problem.

With this PR we
- add a facility in the mock identity provider to simulate the behaviour where the IdP terminates the login flow abnormally by calling the callback URI with `error` and `error_description` parameters
- add a step to the E2E machinery which takes advantage of this, and use it to test that we:
- keep simplifying the handling of exceptions, notably by removing getParams() in filters and their specs
- make the unit tests more discriminating of the exception handling logic, removing a few mocks along the way
- remove checkCode from OidcClientUtilsService, and let oidc-client raise an exception
- upon catching this exception **display the actual error in the error page**, logged (and other usual behaviours for an exception)


back/libs/oidc-client/src/exceptions/oidc-client-missing-code.exception.ts
```diff
@@ -1,15 +0,0 @@
...
-
-export class OidcClientMissingCodeException extends OidcClientBaseException {
-  public code = ErrorCode.MISSING_CODE;
-  public documentation =
-    "La requête reçue au retour du FI n'est pas valide (pas de code d'autorisation), recommencer la cinématique depuis le FS. Si le problème persiste, contacter le support N3";
-  public error = 'server_error';
-  public error_description =
-    'authentication aborted due to a technical error on the authorization server';
-  public http_status_code = HttpStatus.BAD_REQUEST;
-  public ui = 'OidcClient.exceptions.oidcClientMissingCode';
```
rdubigny, line 14: nitpick: we could also remove the corresponding entry in back/apps/core-fca/src/exceptions/error-messages.ts

back/libs/oidc-client/src/exceptions/oidc-client-missing-code.exception.ts
```diff
@@ -1,15 +0,0 @@
-import { HttpStatus } from '@nestjs/common';
-
-import { ErrorCode } from '../enums';
-import { OidcClientBaseException } from './oidc-client-base.exception';
-
-export class OidcClientMissingCodeException extends OidcClientBaseException {
-  public code = ErrorCode.MISSING_CODE;
```
rdubigny, line 7: nitpick: we could also remove this enum (and any non used enum) in back/libs/oidc-client/src/enums

rdubigny (approved):

PR #221, opened 2025-07-15
🐛 fix(oidc-provider): resolve EventBus dependency issue in module test

douglasduteil:
<div align=center></div>

---

Fixes failing test in `oidc-provider.module.spec.ts` by overriding the `OidcProviderRenderedJsonExceptionFilter` with a mock implementation to avoid EventBus dependency resolution error during module testing.

The test now properly validates the module compilation without requiring the full dependency tree.


douglasduteil-admin:
[Refacto : désécriture du ressource serveur](https://trello.com/c/j8Q0yrHH/729-refacto-d%C3%A9s%C3%A9criture-du-ressource-serveur)

PR #226, opened 2025-07-15
Upgrade Nest from v6 to v8

BenoitSerrano:


rdubigny:
[Dette technique : clean de l'admin (1/X)](https://trello.com/c/Os6g46g9/748-dette-technique-clean-de-ladmin-1-x)

admin/shared/form/interceptor/form-errors.interceptor.ts
```diff
@@ -35,27 +38,41 @@ export class FormErrorsInterceptor implements NestInterceptor {
...
-        return of(parsedError).pipe(
-          map(validationErrors => {
-            return validationErrors.reduce(
-              (acc, currentValidationError) => ({
-                ...acc,
-                [currentValidationError.property]:
-                  currentValidationError.constraints,
+        let errorMessage;
+        try {
+          errorMessage = JSON.parse(error.message);
```
rdubigny, line 43: question: is the JSON.parse still needed ?

admin/shared/form/interceptor/form-errors.interceptor.ts
```diff
@@ -35,27 +38,41 @@ export class FormErrorsInterceptor implements NestInterceptor {
...
+        } catch (e) {
+          return next.handle();
+        }
+
+        const formValidationErrors = plainToClass(FormValidationErrorsDto, {
+          message: errorMessage,
+        });
```
rdubigny, line 50: suggestion: `plainToClass(FormValidationErrorsDto, error)`

admin/shared/form/interceptor/form-errors.interceptor.ts
```diff
@@ -35,27 +38,41 @@ export class FormErrorsInterceptor implements NestInterceptor {
...
+
+        return from(validate(formValidationErrors)).pipe(
```
rdubigny, line 52: question: why are you validating the object within a rxjs event queue instead of using classic imperative programming ?

admin/shared/form/interceptor/form-errors.interceptor.ts
BenoitSerrano: I would love to, bruh - and this will answer your previous comment
Without the parsing it crashed with the exception "oops error.message is supposed to be an array, a string was given"
This solution, though inelegant, is the shortest one I came up with

admin/shared/form/interceptor/form-errors.interceptor.ts
BenoitSerrano: cf. below

rdubigny (approved):

PR #229, opened 2025-07-16
💚 Remove the incorrectly configured admin backup in dev

gdarquie:
Actually, I have some doubts about what I've done, I continue to investigate. Please DO NOT MERGE :pray:


rdubigny (approved):

gdarquie:
This PR was a bad idea—we should retain a way to reset the PostgreSQL database, and the backup operations here are used for that purpose.

This other PR is a better way, I think, to upgrade the hooks: https://github.com/proconnect-gouv/federation/pull/233

PR #233, opened 2025-07-17
🎨 Clean _up hooks

gdarquie:


gdarquie:
[Nettoyage des hooks de _up](https://trello.com/c/QX7oybls/757-nettoyage-des-hooks-de-up)

PR #234, opened 2025-07-17
♻️  complete rewriting of logger-legacy

rdubigny:
There was much dead code in this lib.

I propose a complete iso-functionnal rewriting of the lib. We only keep the function in use.


back/libs/logger-legacy/src/services/pino.service.ts
```diff
@@ -1,53 +0,0 @@
...
-
-import { LoggerConfig as LoggerLegacyConfig } from '../dto';
-import { LoggerLevelCodes } from '../enum';
-import { pinoLevelsMap } from '../log-maps.map';
-
-/**
- * Formatter for pino library
- * @see https://github.com/pinojs/pino/blob/master/docs/api.md#formatters-object
- */
-export function formatLevel(label: string, _number: number): { level: string } {
```
gdarquie, line 15: question: I don't really understand that part? why we needed a "custom" pino service before?

What I understand here, it's that we needed a custom stream build with `getDestinationForLog`?

gdarquie (approved):

PR #237, opened 2025-07-17
🚚 move core-tracking-service to tracking module

rdubigny:
We have no use of core-tracking-service being generic anymore.

I propose to move it to the tracking module where it is used.


back/libs/core/src/core.module.ts
```diff
@@ -113,7 +110,7 @@ export class CoreModule {
       exports: [
         CoreAccountService,
         CoreOidcProviderConfigAppService,
-        trackingModule,
+        TrackingModule,
```
gdarquie, line 113: Not sure if we still need to export the TrackingModule now, since it's just the standard module without any injection.

rdubigny: this is deleted in #238

gdarquie (approved):

PR #239, opened 2025-07-18
🔨 Add a dks command for cleaning node modules and dist

gdarquie:
This command helps clean the Node.js stack by removing node_modules folders and built JavaScript files. It is particularly helpful when working with the Admin which is composed of multiple apps:
- the parent admin app
- shared
- fc-exploitation

Each of these apps has its own node_modules folder.


gdarquie:
[Simplification du lancement des stacks](https://trello.com/c/W5r7Q6Qw/710-simplification-du-lancement-des-stacks)

PR #287, opened 2025-07-29
🧱 Add a new stack for testing prod docker image

gdarquie:
For now, this is still an initial proof of concept (POC) that allows launching a minimal development stack using the same Docker package as in production.
The version is currently hardcoded, and only the core uses a GitHub package.

After launching the stack with the `dks switch prod command`, the application still needs to be started manually.

To start the app:

`dks compose exec core-prod pm2-runtime /etc/pm2/app.json`


gdarquie:
[Faire un POC d'une stack qui utilise les images de prod](https://trello.com/c/OpmBHNav/764-faire-un-poc-dune-stack-qui-utilise-les-images-de-prod)

PR #295, opened 2025-07-31
👷 add more artifacts

rdubigny:


.github/workflows/ci-tests.yml
```diff
@@ -146,8 +146,10 @@ jobs:
         with:
           name: cypress-snapshots-artifacts
           path: |
+            quality/fca/cypress/screenshots
+            quality/fca/cypress/videos
```
gdarquie, line 150: question: sure about video because we `yarn test:low:snapshot -c video=false`?

.github/workflows/ci-tests.yml
```diff
@@ -146,8 +146,10 @@ jobs:
...
+            quality/fca/cypress/logs
             quality/fca/cypress/reports/cucumber/html-report
-            quality/fca/cypress/snapshots/diff
```
gdarquie, line 150: question: why no more snapshots?

image

PR #302, opened 2025-08-01
🔨 In dev install node module in Dockerfile and no more during _up

gdarquie:


docker/bash/commands/docker.sh
```diff
@@ -1,20 +1,25 @@
...
 
   echo -e "${ancestor_containers}\n${pattern_containers}" | sort | uniq
 }
 
 # Find which nodejs containers are running and store it into $NODEJS_CONTAINERS
-_get_running_containers() {
-  local raw_nodejs_containers=$(_get_node_containers)
-  local raw_all_containers=$(docker ps --format '{{.Names}}')
+_get_node_containers_to_start() {
+  local ancestor_containers=$(docker ps --format '{{.Names}}' -f ancestor=pc-nodejs)
```
BenoitSerrano, line 13: As not everyone in the team is fluent in bash (I'm not hihi), could you write extra comments on these lines to help understand how this function works? And/or rename the variables to be super explicit about what they're supposed to contain?

PR #303, opened 2025-08-04
Flatten admin with gaetan

BenoitSerrano:


gdarquie:
[Admin : Simplification des dossiers](https://trello.com/c/i3dEbW5n/770-admin-simplification-des-dossiers)

PR #362, opened 2025-08-13
✨ send extra params should remember me

rebeccadumazert:


back/apps/core-fca/src/dto/redirect-to-idp.dto.ts
```diff
@@ -10,4 +17,9 @@ export class RedirectToIdp extends CrsfToken {
   @IsOptional()
   @IsAscii()
   readonly identityProviderUid?: string;
+
+  @IsBoolean()
+  @IsOptional()
+  @Transform(({ value }) => value === 'on' || value === true)
```
rdubigny, line 23: The value comes from a form on our side. I expect the transformation to have only one comparaison if any.

rdubigny (commented):
shouldRememberMe (and it's case variants) can be renamed to rememberMe everywhere.

gdarquie (approved):

back/apps/core-fca/src/services/core-fca.service.ts
```diff
@@ -73,6 +74,7 @@ export class CoreFcaService {
       login_hint,
       sp_id: spId,
       sp_name: spName,
+      remember_me: rememberMe || false,
```
rdubigny, line 77: issue: in a multi FI scénario we send remember_me false every time

rdubigny (commented):
I have found an issue with this implementation. We need to discuss this.

back/apps/core-fca/src/services/core-fca.service.ts
rdubigny: fixed in #445

PR #363, opened 2025-08-13
Upgrade last dependencies admin

BenoitSerrano:


rdubigny (commented):
@BenoitSerrano I do not want to maintain this kind of functions in the codebase. Plz keep lodash.

PR #365, opened 2025-08-14
Simplify interfaces identity provider

BenoitSerrano:


admin/src/fqdn-to-provider/fqdn-to-provider.service.ts
```diff
@@ -35,119 +37,95 @@ export class FqdnToProviderService {
...
   }
 
   async saveFqdnsProvider(
     identityProviderUid: string,
-    fqdns: Array<Pick<IFqdnToProvider, 'fqdn' | 'acceptsDefaultIdp'>>,
+    fqdns: string[],
   ): Promise<void> {
-    fqdns.filter(Boolean).forEach(async (fqdn) => {
+    const filteredFqdns = fqdns.filter(Boolean);
+    for (let index = 0; index < filteredFqdns.length; index++) {
```
jonathanperret, line 71:
```suggestion
    for (const fqdn of filteredFqdns) {
```

gdarquie (approved):
Hello,

After reviewing this PR, I have some feedback:

- 51 commits is too many for this kind of work. I think it should be possible to regroup them into fewer than 10 commits.
- Commit messages like "rename variable to be more explicit", "rename variable", "rename variables to be more explicit", etc. are too generic to be useful later. A better example would be something like "[Remove working_dir overrides](https://github.com/proconnect-gouv/federation/commit/abfd922872f25d83bac1bb0d4cdd13f7abe7a89b)", which is more specific. This point is related to the previous one: with fewer, more meaningful commits, the commit history would be clearer.
- It might have been better to split this PR into multiple parts. For example, this [commit](https://github.com/proconnect-gouv/federation/pull/365/commits/0fcface7c727215d0e919471982547e36ebe5706) which resolves the bug described in this [ticket](https://trello.com/c/St1rq5VH/789-freeze-de-proconnect-suite-%C3%A0-une-modification-dans-ladmin)
, could have been extracted into a separate PR. Also, though I’m not entirely sure, perhaps the logic for the FQDN could have been extracted first.

admin/src/fqdn-to-provider/fqdn-to-provider.service.ts
gdarquie: Applied in https://github.com/proconnect-gouv/federation/pull/424/files

admin/src/fqdn-to-provider/fqdn-to-provider.service.ts
```diff
@@ -35,119 +37,95 @@ export class FqdnToProviderService {
...
-    const fqdnsToAdd: Array<
-      Pick<IFqdnToProvider, 'fqdn' | 'acceptsDefaultIdp'>
-    > = fqdns
-      .filter(
-        (fqdn) =>
-          !providerFqdns.find((fqdnToProvider) => fqdnToProvider.fqdn === fqdn),
-      )
-      .map((fqdn) => ({ fqdn, acceptsDefaultIdp: true }));
+    const fqdnsToAdd = fqdns.filter(
+      (fqdnToAddFqdn) =>
```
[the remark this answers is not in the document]
gdarquie: Applied in https://github.com/proconnect-gouv/federation/pull/424/files

PR #385, opened 2025-08-20
🔧 Remove cache for apk add in admin Dockerfile

gdarquie:


gdarquie:
[Supprimer le cache apk des images docker](https://trello.com/c/G85U8LxA/779-supprimer-le-cache-apk-des-images-docker)

jonathanperret (approved):
I think `--no-cache` here is OK for consistency, but note that none of the claimed motivations exist here:
- this is an intermediate stage so the image size is irrelevant;
- the parent image (`node:22.13.0-alpine`) does not contain an existing cache so there is no risk of using an old cache.

PR #201, opened 2025-07-11
🐳 refactor(docker): use extends pattern to reduce configuration duplication

douglasduteil:
<div align=center></div>

---

## Problem
Docker Compose configurations had ~70% duplication across Node.js and Redis services, making maintenance difficult and error-prone.

## Solution
Introduced extends pattern with base service definitions to eliminate configuration duplication:

- Added base service definitions in shared.yml for common patterns
- Refactored Node.js services to extend base-nodejs-app
- Created specialized base services for core, FSA, FIA, DPA, and RIE patterns  
- Refactored Redis services to use base-redis pattern
- Fixed inconsistent volume path in fsa1-low service
- Reduced ~70% configuration duplication across services

PR #388, opened 2025-08-21
🧑 Use extends in compose core files

gdarquie:
This PR is an update of https://github.com/proconnect-gouv/federation/pull/201 from @douglasduteil with some minor changes.

Compose does not permit partially overriding a list, so for env_files we need to explicitly specify all the files. However, this is not needed for depends_on.


gdarquie:
[Utilisation d'extends dans docker compose](https://trello.com/c/P9qvGw4l/797-utilisation-dextends-dans-docker-compose)

PR #394, opened 2025-08-22
🧑 Use extends for redis in compose files

gdarquie:


gdarquie:
> I hope we can remove the need for a redis cluster in dev one day 📿

Have looked at that https://github.com/proconnect-gouv/federation/pull/395/commits/17d7e9eba51bcc5f047726e0b1cec69842770b1a ? 

We still use redis-pwd but bye bye `redis-cluster-master` or `redis-cluster-replica`

PR #398, opened 2025-08-25
♻️ Remove command override from Compose services

gdarquie:
Using the existing NESTJS_INSTANCE environment
variable lets all those containers share a command defined in the common Dockerfile.


gdarquie:
[Déplacer les starts de dks dans les Dockerfiles](https://trello.com/c/RZ5PNvsJ/802-d%C3%A9placer-les-starts-de-dks-dans-les-dockerfiles)

PR #399, opened 2025-08-25
🚚 Replace environment variables with relative paths in Compose files

gdarquie:
From https://github.com/proconnect-gouv/federation/commits?author=jonathanperret

Replace environment variables with relative paths in Compose files This makes the Compose stack more self-contained, less dependent on the `docker-stack` wrapper.

Note that `_clean_pc_dist` in `node.sh` was the last user of `VOLUMES_DIR` but was itself unused so I removed it, along with its `_detect_instances` and `_get_env` dependencies.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -123,12 +123,12 @@ services:
     depends_on:
       - <...any_required_service...>
     volumes:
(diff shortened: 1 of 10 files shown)


douglasduteil:
Me Like :heart:

PR #412, opened 2025-08-26
✅ Check in CI the RIE error when there is  no idp found

gdarquie:


gdarquie:
[Ajouter une stack RIE](https://trello.com/c/KucT3zUA/803-ajouter-une-stack-rie)

PR #413, opened 2025-08-26
🔥 Remove working_dir overrides

gdarquie:
The working directory is already specified in the Dockerfile.


gdarquie:
[Supprimer des compose les working dir](https://trello.com/c/UxV2uBZ1/806-supprimer-des-compose-les-working-dir)

PR #414, opened 2025-08-26
🔨 Streamline dev Docker images

gdarquie:
Yarn will store its cache by default in `/usr/local/share/.cache`.

Instead of keeping that in the final image, we move it to a cache mount which gives it an opportunity to be reused by the next build.

We also make /tmp a tmpfs mount because yarn dumps stuff there as well.


gdarquie:
[Sauvegarder le cache yarn dans un volume dans la stack de dev](https://trello.com/c/ZfAYmqXM/807-sauvegarder-le-cache-yarn-dans-un-volume-dans-la-stack-de-dev)

PR #417, opened 2025-08-27
🏗  Delete core-fca instance

gdarquie:


gdarquie:
[Supprimer les instances](https://trello.com/c/IwX6jkRe/776-supprimer-les-instances)

PR #418, opened 2025-08-27
✅ Add a stack for testing core and admin github packages

gdarquie:


gdarquie:
[Tester les images buildées de prod via la stack de dev](https://trello.com/c/gA9nWiiX/783-tester-les-images-build%C3%A9es-de-prod-via-la-stack-de-dev)

PR #500, opened 2025-09-22
♻️  remove session duplication in authorize

rdubigny:
We are not sure why the session was duplicated.
It seems intended as a session key rotation to mitigate cookie theft attacks, but the previous key does not appear to be revoked.

Such protection is relevant when elevating privileges within a session or when revoking a session using the same key, but that's not the case in the authorize endpoint. It still applies in the callback endpoint.

This duplication would cause issues when refactoring the prompt to matche the Panva API.

I propose removing it now to make the code clrearer.


gdarquie:
question: why isn’t the PR comment included in the body of the unique commit?

It seems there are a couple of small typos in it — I found two: “matche” and “clrearer”.

PR #516, opened 2025-09-24
Reconcile identities from PCI with a new IdP

Morendil:


Morendil:
Closing in favor of #517

PR #517, opened 2025-09-24
On connection, store email in FcaAccount.idpIdentityKeys

Morendil:


back/libs/account-fca/src/services/account-fca.service.ts
```diff
@@ -54,8 +56,8 @@ export class AccountFcaService {
       throw new CoreFcaAgentAccountBlockedException();
     }
 
-    if (!some(account.idpIdentityKeys, idpAgentKeys)) {
-      account.idpIdentityKeys.push(idpAgentKeys);
+    if (!some(account.idpIdentityKeys, idpFullKeys)) {
```
rdubigny, line 59: issue: the algorithm should be:

if an idpIdentityKey exist with these specific idpUid and idpSub
then we update the email of the corresponding idpIdentityKeys
else we push the entire idpFullKeys in the list.

The problem here is that idpMail is used in the match.

rdubigny (changes requested):

rdubigny:
[Lorsqu'un utilisateur PCI, migre vers un autre FI en conservant la même adresse mail, il conserve son sub fédération](https://trello.com/c/tTwxG1QW/736-lorsquun-utilisateur-pci-migre-vers-un-autre-fi-en-conservant-la-m%C3%AAme-adresse-mail-il-conserve-son-sub-f%C3%A9d%C3%A9ration)

PR #523, opened 2025-09-29
feat(back): use Error.cause

douglasduteil:
<div align=center>  </div>

---

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause


douglasduteil:
Outdated

PR #526, opened 2025-09-29
Use fqdns instead of fqdn to provider

BenoitSerrano:


gdarquie (approved):

back/apps/core-fca-low/src/services/core-fca-fqdn.service.ts
```diff
@@ -30,9 +31,7 @@ export class CoreFcaFqdnService {
       };
     }
 
-    const uniqueDuplicateFreeIdpUids = uniq(
-      idpsByFqdn.map(({ identityProvider }) => identityProvider),
-    );
+    const uniqueDuplicateFreeIdpUids = uniq(idpsByFqdn.map(({ uid }) => uid));
```
jonathanperret, line 34: Not only are `unique` and `DuplicateFree` redundant (ironically); but also, `getIdpsByFqdn` can no longer return duplicate IdPs since it returns a subset of the IdP list.

```suggestion
    const identityProviderIds = idpsByFqdn.map(({ uid }) => uid);
```

PR #543, opened 2025-10-01
🔥 Remove dks log

gdarquie:
`dks log` used pm2 logs,
but since we have removed pm2 in federation,
this command is no longer needed.


jonathanperret (approved):
The command is not just no longer needed, it's plain broken! Note that there's one remaining occurrence of pm2 in wait.sh.

douglasduteil (approved):

PR #1087, opened 2026-04-08
🐋 (back): make the image use the /api/v2/livez as default healthcheck

douglasduteil:
**Problem**

We need to manually set the healthcheck on the image.

**Proposal**

Add a default livez healthcheck to the image.


douglasduteil:
I know `https` requests from inside the container is strange...
For me, it shows that we might want to consider opening HTTP port too.
See https://docs.nestjs.com/faq/multiple-servers#multiple-simultaneous-servers

Or we just remove the hardcoded httpsOptions entirely...

BenoitSerrano (approved):

PR #1121, opened 2026-04-20
🩺 (fca-low): exclude the hyyyperbridge from readiness when flagged

douglasduteil:
**Problem**

The hyyyperbridge is an optional feature that can be disabled in some cases.

**Proposal**

Use the flag to exclude the hyyyperbridge from readiness checks.

### DIFF
--- a/back/apps/core-fca-low/src/controllers/health.controller.spec.ts
+++ b/back/apps/core-fca-low/src/controllers/health.controller.spec.ts
@@ -1,4 +1,5 @@
 import { ApiEntrepriseService } from "@fc/api-entreprise";
+import { ConfigService } from "@fc/config";
 import { RedisService } from "@fc/redis";
 import { HttpStatus } from "@nestjs/common";
 import { Test, TestingModule } from "@nestjs/testing";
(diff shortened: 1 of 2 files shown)


rebeccadumazert (approved):

back/apps/core-fca-low/src/controllers/health.controller.ts
```diff
@@ -53,6 +56,10 @@ export class HealthController {
       await this.apiEntreprise.getOrganizationBySiret("13002526500013");
     },
     [ExcludeTarget.Hyyyperbridge]: async () => {
+      const { bypassHybridgeInternet } =
+        this.config.get<OidcClientConfig>("OidcClient");
+      if (!bypassHybridgeInternet) return "bypass";
```
BenoitSerrano, line 61: I'm not sure I understand the logic here: if "bypass" is falsy, then we return "bypass" ?

back/apps/core-fca-low/src/controllers/health.controller.spec.ts
```diff
@@ -151,6 +160,16 @@ describe("HealthController", () => {
       expect(result).toBe("error");
     });
 
+    it('should return "ok" when Hyyyperbridge is bypassed', async () => {
+      configServiceMock.get.mockReturnValue({ bypassHybridgeInternet: false });
```
BenoitSerrano, line 164: hyyyperbridge is bypassed === bypassHybridgeInternet: false ?

douglasduteil:
Here is the cause : https://github.com/proconnect-gouv/federation/blob/38140379b93f0d516c0f4e136f07967edca5c354/back/libs/oidc-client/src/services/oidc-client.service.ts#L252
@BenoitSerrano talk to @rdubigny about it... I'm confused too

PR #1122, opened 2026-04-20
🐛 do no assign acr field if acrs not recognized

BenoitSerrano:
**Problem**
Entra ID Identity Providers sometimes return `acrs` values that cannot be interpreted.

**Proposal**
Fallback to a default value of `eidas1` when the `acrs` field cannot be parsed. ``

### DIFF
--- a/back/libs/oidc-client/src/services/oidc-client.service.ts
+++ b/back/libs/oidc-client/src/services/oidc-client.service.ts
@@ -22,7 +22,7 @@ import { ClientProxy } from "@nestjs/microservices";
 import { plainToInstance } from "class-transformer";
 import { validate } from "class-validator";
 import { Request } from "express";
-import { chain, cloneDeep, filter, isObject } from "lodash";
+import { chain, cloneDeep, isObject, map } from "lodash";
 import {
   authorizationCodeGrant,
   AuthorizationResponseError,
@@ -73,6 +73,14 @@ export class OidcClientService {
     "phone",
     "chorusdt",
   ].join(" ");
+
+  entraIdAcrValuesMapping = {
+    c1: "eidas1",
+    c2: "eidas2",
+    c3: "eidas3",
+  };
+
+  prioritizedAcrValues = ["eidas3", "eidas2", "eidas1"];
 
   constructor(
     private readonly config: ConfigService,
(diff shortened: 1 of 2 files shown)


BenoitSerrano:
[Correction de la gestion de EntraID pour SMICA et OBD](https://trello.com/c/ZyGUv7l1/1335-correction-de-la-gestion-de-entraid-pour-smica-et-obd)

back/libs/oidc-client/src/services/oidc-client.service.ts
```diff
@@ -74,6 +74,12 @@ export class OidcClientService {
     "chorusdt",
   ].join(" ");
 
+  entraIdEidasLevelMapping = {
```
rdubigny: nitpick: `entraIdAcrValuesMapping`, eidasLevel refers more to the EU legislation terminology. Internaly, we are dealing with acr values.

back/libs/oidc-client/src/services/oidc-client.service.ts
```diff
@@ -427,8 +429,13 @@ export class OidcClientService {
       // This behavior is specific to MS Entra ID's authentication contexts
       // We consider only values of the form c1, c2, c3, mutually exclusive,
       // and map them to equivalent eidas levels
-      const acrs = filter(token.claims.acrs, (level) => level.startsWith("c"));
-      acr = acrs.toString().replace("c", "eidas");
+      const newAcrs = map(
```
rdubigny: nitpick: `newAnythin` are like variable name called `data` they do not designate anything specific and should be avoided. I suggest `proConnectAcrs`.

back/libs/oidc-client/src/services/oidc-client.service.ts
```diff
@@ -427,8 +429,13 @@ export class OidcClientService {
...
+        token.claims.acrs,
+        (level) => this.entraIdEidasLevelMapping[level],
+      ).filter(Boolean);
+      if (!!newAcrs.length) {
+        acr = newAcrs.join(" ");
```
rdubigny: issue: according to [the specification](https://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken):

> acr: OPTIONAL. Authentication Context Class Reference. String specifying an Authentication Context Class Reference value that identifies the Authentication Context Class that the authentication performed satisfied.

Therefore, this value should not be a space-separated string.

rdubigny (changes requested):

rdubigny (approved):
brilliant!

PR #1139, opened 2026-04-22
🔊 add logs for acr_values

BenoitSerrano:
**Problem**
The `acr_values` field is deprecated and should no longer be used, but removing it outright could introduce regressions for existing users.

**Proposal**
Add logging to track occurrences of this parameter in order to assess usage before deprecating it safely.


BenoitSerrano:
[Supprimer la gestion legacy du acr_values=eidas1 (sans bloquer les FS qui l'utilisent)](https://trello.com/c/TDQmeBPh/1311-supprimer-la-gestion-legacy-du-acrvalueseidas1-sans-bloquer-les-fs-qui-lutilisent)

rdubigny (approved):

PR #1143, opened 2026-04-27
🐛 oidc-client: wire customFetch on non-discovery Configuration

douglasduteil:
**Problem**

For IdPs that don't use discovery, `Configuration` was created without `[customFetch]`, so `authorizationCodeGrant` and other token requests bypassed the hyyyperbridge even when `idp.useTheHyyyperbridge` was set.

**Proposal**

Assign `config[customFetch]` after the non-discovery `new Configuration()` call, mirroring the same binding used in the discovery branch. 

see https://github.com/panva/openid-client/blob/v6.8.4/docs/variables/customFetch.md


arnaud-robin (approved):

rdubigny (approved):
Brilliant

PR #1123, opened 2026-04-20
Revert "🗑️ delete api health test"

douglasduteil:
This test should work.
@BenoitSerrano why did we removed it ? Did I miss a `@tag` for the integration env ?

Reverts proconnect-gouv/federation#1120
Introduced in proconnect-gouv/federation#1117

PR #1154, opened 2026-05-07
✨ (core-fca-low): support multiple excludes on readyz

douglasduteil:
**Problem**

The `/readyz` endpoint only accepted a single `?exclude=` value, making it impossible to skip more than one check per request.

**Proposal**

Switch to `ParseArrayPipe` so repeated `?exclude=` params are collected into an array. 
Rename `ExcludeTarget` → `CheckTarget` for clarity (used by both the exclude list and the per-check route).

Introduced in https://github.com/proconnect-gouv/federation/pull/1117
Reverts https://github.com/proconnect-gouv/federation/pull/1120
Closes #1123

### DIFF
--- a/back/apps/core-fca-low/src/controllers/health.controller.ts
+++ b/back/apps/core-fca-low/src/controllers/health.controller.ts
@@ -9,6 +9,7 @@ import {
   HttpStatus,
   Inject,
   Param,
+  ParseArrayPipe,
   ParseEnumPipe,
   Query,
   Res,
@@ -20,7 +21,7 @@ import { Connection } from "mongoose";
 import { firstValueFrom, timeout } from "rxjs";
 import { Routes } from "../enums";
 
-export enum ExcludeTarget {
+export enum CheckTarget {
   ApiEntreprise = "api-entreprise",
   Hyyyperbridge = "hyyyperbridge",
   MongoDB = "mongodb",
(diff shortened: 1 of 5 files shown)


back/apps/core-fca-low/src/controllers/health.controller.ts
```diff
@@ -79,10 +80,10 @@ export class HealthController {
   async readyz(
     @Res({ passthrough: true }) res: Response,
     @Query("verbose") verbose?: string,
-    @Query("exclude", new ParseEnumPipe(ExcludeTarget, { optional: true }))
-    exclude?: ExcludeTarget,
+    @Query("exclude", new ParseArrayPipe({ items: String, optional: true }))
+    exclude: CheckTarget[] = [],
```
rdubigny, line 84: thoughts: You are not testing these parameters against a whitelist of values anymore. This could lead to unexpected usage.

nitpick: Exclude is actually an array of string.

rdubigny (approved):
Thought: This controller is overly complicated. Its only usage is to check whether we can connect to Redis and MongoDB, so it should do these two things in about ten lines.

It opens too many attack vectors for very little gain.

back/apps/core-fca-low/src/controllers/health.controller.ts
douglasduteil: > thoughts: You are not testing these parameters against a whitelist of values anymore. This could lead to unexpected usage.

I do not consider a list of excluded string as critical as this only impact is for it to skip a check. Hence string outside of the whitelist will silently do nothing. 

```
exclude=foo,bar => not impact as the checks list is determined by the code.
```

The only "unexpected usage" is pure human mistakes I consider low risk.

> nitpick: Exclude is actually an array of string.

Yep want me to downgrade the type ?

PR #1157, opened 2026-05-11
🐛 Only display red border on admin in prod

gdarquie:
## Context

image

Actually, on Kubernetes, all admin environments — including preprod — display a red border. However, the red border is supposed to be displayed only in production.

I also realized that there were two different ways to add borders to the admin, resulting in two different border styles.

## Proposition

I propose removing one of the two ways to add a border, keeping the simplest approach in a first commit.

Then, I’ll update the logic so the red border is displayed only when the `IS_PRODUCTION` environment variable is set to `"true"`. With the previous code, the border turned red whenever `IS_PRODUCTION` existed, regardless of its value.

Actually, the `isProduction` concept is only used to display the red border, so we rename it to something more precise: `hasRedBorder`.

PS: The following code in `/admin/src/meta/locals.interceptor.ts` maps `locals.IS_PRODUCTION` to `isProduction`:

```js
    res.locals.IS_PRODUCTION = appConfig.isProduction;
```


BenoitSerrano (approved):

douglasduteil (approved):
LGTM haha

rdubigny (approved):

PR #1158, opened 2026-05-11
✨ use roles value to prevent non-public user to access SP

BenoitSerrano:
**Problem**
Private sector users are incorrectly blocked due to the legacy `is_service_public` calculation.

**Proposal**
Block users based on the `roles` field as computed from the SIRET instead of relying on the legacy `is_service_public` logic.


douglasduteil (approved):
Looks cool :)

back/apps/core-fca-low/src/controllers/interaction.controller.ts
```diff
@@ -253,25 +252,10 @@ export class InteractionController {
     }
 
     const { type: spType } = await this.serviceProvider.getById(spId);
-    // is_service_public field is only provided by ProConnect Identité
-    // any identity without an is_service_public field is considered to be from the public sector
-    const isPrivateSectorIdentity = idpIdentity?.is_service_public === false;
     const isRoleAgentPublic = !roles || roles.includes("agent_public");
```
rdubigny: question: should we keep `!roles` ? It is quite unexpected that having no roles define a public servant.

back/apps/core-fca-low/src/dto/identity-from-idp.dto.ts
```diff
@@ -72,10 +71,6 @@ export class IdentityFromIdpDto {
   @IsOptional()
   "chorusdt:matricule"?: string;
 
-  @IsBoolean()
-  @IsOptional()
-  is_service_public?: boolean;
```
rdubigny, line 77: thought: As PCI still send it, is_service_public will be added to the custom claim.

back/apps/mock-identity-provider-fca-low/src/index.ejs
```diff
@@ -1,79 +1,128 @@
 <!doctype html>
 <html lang="fr-FR">
-<head>
+  <head>
```
rdubigny: nitpick: linting seems out of the scope of this PR. It is hard to see if there is any changes here.

back/apps/core-fca-low/src/dto/identity-for-sp.dto.ts
```diff
@@ -35,6 +36,10 @@ export class IdentityForSpDto extends IdentityFromIdpDto {
   @IsString()
   idp_acr: string;
 
+  @IsBoolean()
+  @IsOptional()
+  is_service_public?: boolean;
```
rdubigny: todo: This was not returned. We should not start returning it. We can do some additional cleaning in the sanitizer.

back/apps/core-fca-low/src/services/identity.sanitizer.ts
```diff
@@ -58,30 +59,42 @@ export class IdentitySanitizer {
...
     const { featureFetchOrganizationData } =
       this.config.get<ApiEntrepriseConfig>("ApiEntreprise");
 
-    if (featureFetchOrganizationData && !!identityForSp.siret) {
+    if (featureFetchOrganizationData) {
       try {
         const cachedOrganization =
           await this.cachedOrganizationService.getCachedOrganizationBySiret(
-            identityFromIdp.siret,
+            identityFromIdp.siret!,
```
rdubigny: issue: at this point `identityForSp.siret` is defined but `identityFromIdp.siret` might not be.

back/apps/core-fca-low/src/services/identity.sanitizer.ts
```diff
@@ -137,32 +150,46 @@ export class IdentitySanitizer {
...
 
-      this.logger.error({
-        msg: `Identity from "${idpId}" is invalid`,
-        code: "invalid-identity-validation",
-        validationErrors: identityValidationErrors,
-      });
-
-      this.logger.debug({
-        code: "invalid-identity-debug",
+      await this.throwExceptionForInvalidIdentity(
```
rdubigny: nitpick: await as no effect here.

rdubigny (changes requested):

rdubigny (approved):

PR #1161, opened 2026-05-12
🚸 add default roles in admin

BenoitSerrano:
**Problem**
The `roles` field currently has to be added manually each time a new Service Provider is created.

**Proposal**
Add the `roles` scope to the default preselected scopes in the admin interface.


BenoitSerrano:
[Ajout de scope roles par défaut dans admin](https://trello.com/c/BstmFjl4/1358-ajout-de-scope-roles-par-d%C3%A9faut-dans-admin)

rdubigny (approved):

PR #1184, opened 2026-05-21
🗑️ get rid of database scopes in admin

BenoitSerrano:
**Problem**
Scopes stored in the database require ongoing maintenance whenever changes are made, without providing any real benefit.

**Proposal**
Store scopes directly in the codebase instead of persisting them in the database.


BenoitSerrano:
[Ajouter les scopes en dur dans le code de l'admin](https://trello.com/c/vQAwyraj/1373-ajouter-les-scopes-en-dur-dans-le-code-de-ladmin)

rdubigny (approved):

PR #1198, opened 2026-05-28
🗑️ Remove BridgeHttpProxyRie app

rdubigny:
**Problem**

The `BridgeHttpProxyRie` app has been replaced by `hyyyperbridge`. The connection through a RabbitMQ queue is now directly embedded in the OIDC client.

We no longer need `BridgeHttpProxyRie`.

**Proposal**

Remove `BridgeHttpProxyRie`.


douglasduteil (approved):
<div align=center>  </div>

PR #1199, opened 2026-06-01
🔊 add log when discrepancy between legacy and new computation of service public

BenoitSerrano:
**Problem**
The new public agent status calculation requires communication and impact assessment before being rolled out to production.

**Proposal**
Log discrepancies between the legacy and the new calculation in production to identify affected users and proactively inform them ahead of the rollout.


back/libs/cached-organization/src/services/cached-organization.service.ts
```diff
@@ -25,6 +30,22 @@ export class CachedOrganizationService {
...
     });
+    const servicePublicInfo = computeServicePublicInfo({
+      cached_categorie_juridique: cachedOrganization.categorieJuridique,
+      cached_etat_administratif: cachedOrganization.etatAdministratif,
+      siret: cachedOrganization.siret,
+    });
+
+    if (servicePublicInfo.isServicePublic !== isServicePublic) {
+      this.logger.warn({
+        code: "cached-organization-service-public-mismatch",
```
rdubigny, line 41: Nitpick: I am not sure “cached” adds any useful meaning to this label.

rdubigny (approved):

back/libs/cached-organization/src/services/cached-organization.service.ts
BenoitSerrano: It is the name of the service, I thought it was the rule to log the complete name of the service

PR #1201, opened 2026-06-03
🎉 Create statisticat for computing stats

gdarquie:
Create a new image for log extraction and stats generation to be used in k8s-proconnect.

Later I will update the temporary tag (test-extract-logs-stats) to the standard pcf-XXX format in k8s-proconnect.


gdarquie:
[Rétablir les stats de connexion](https://trello.com/c/q05kLjfW/1352-r%C3%A9tablir-les-stats-de-connexion)

PR #1216, opened 2026-06-05
🔧 docker: inline MONGO_VERSION for dependabot

douglasduteil:
**Problem**

The mongo version is declared twice (Dockerfile `ARG` default and compose `args`) and dependabot's docker-compose updater only tracks base-image references it can parse from a Dockerfile `FROM` line, not values passed via build args. As a result, mongo upgrades still need a hand-written PR.

**Proposal**

Inline the version into `FROM mongo:5.0.23` and drop the build-arg indirection, so dependabot detects it as a base image and opens the upgrade PR itself.

\following https://github.com/proconnect-gouv/proconnect-espace-partenaires/pull/345#issue-comment-box

### DIFF
--- a/docker/builds/mongodb/Dockerfile
+++ b/docker/builds/mongodb/Dockerfile
@@ -1,5 +1,4 @@
-ARG MONGO_VERSION=5.0.23
-FROM mongo:${MONGO_VERSION}
+FROM mongo:5.0.23
 
 RUN openssl rand -base64 756 > /mongo-key-file
 COPY ./volumes/ssl/docker-stack-ca.crt /usr/local/share/ca-certificates/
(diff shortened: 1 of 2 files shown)


docker/builds/mongodb/Dockerfile
```diff
@@ -1,5 +1,4 @@
-ARG MONGO_VERSION=5.0.23
-FROM mongo:${MONGO_VERSION}
+FROM mongo:5.0.23
```
rdubigny, line 1: note: version is now 8.0.3 since https://github.com/cloud-gouv/k8s-proconnect/commit/7a9a3ef810bfe39246601c30798aeb3a5f046b9c 

Did we made an upgrade of 3 major versions without noticing it?

cc. @gdarquie @kmasseix-dinum

kmasseix-dinum: Probable.

rdubigny (approved):

docker/builds/mongodb/Dockerfile
rdubigny: seen with @gdarquie, this change was intended. The downgrade to v5 on kube was estimated more risky.

PR #1230, opened 2026-06-10
🚫 ci: prevent unnecessary test runs on main merges

BenoitSerrano:
**Problem**
Tests are systematically triggered on merges to `main`, even though they are not meaningful in that context.

**Proposal**
Filter workflow execution to prevent unnecessary test runs when merging into `main`.


rdubigny (commented):
no

douglasduteil (changes requested):
Nope. 
The truth is the default branch

PR #1243, opened 2026-06-12
🏷️ Update email field label on SP edition page

rdubigny:
**Problem**

The SP email field is labeled as `Email(s) de contact (un par ligne)` on the SP edition page.

This field is not used by Federation, as it does not appear in the SP schema.  Its only purpose is to grant access to the corresponding user in the Espace Partenaires application.

Furthermore, if an admin enters more than one email separated by new lines, as suggested by the current label, the corresponding SP will no longer be editable by any user with one of these email addresses, even though we might expect multiple users to have access.

**Proposal**

Update the label to `Email propriétaire dans l’Espace Partenaires` to describe the field more accurately.

**Boy scout**

Also remove password manager autocompletion on the admin user creation page to avoid TOTP autofill interfering with the username field.


BenoitSerrano (approved):
Should we add a migration that loops over the current values in the database and picks the first email whenever the fields contains several?

rdubigny:
> Should we add a migration that loops over the current values in the database and picks the first email whenever the fields contains several?

I do not think this should be necessary, since any administrator can update this field with a unique email address. We should clean it up as we go.

PR #1255, opened 2026-06-16
🐳 (docker): add redis-sentinel

douglasduteil:
image

---

**Problem**

Local dev has only a plain `redis-pwd` (port 6379). Production runs a
1+1+3 sentinel cluster for HA, but there is no way to exercise the ioredis
sentinel code path locally.

**Proposal**

`dks switch redis-sentinel` — local replica of the production HA topology
so the ioredis [sentinel code path][ioredis-sentinel] can be exercised
without deploying to k8s.

[Redis Sentinel][redis-sentinel] is Redis's built-in HA mechanism.
Sentinels are independent processes that monitor a master/replica pair
and coordinate a failover — promoting the replica to master — when the
master becomes unreachable. A quorum (the minimum number of sentinels
that must agree the master is down before acting) prevents false
failovers from transient network issues. At least 3 sentinels are
required so quorum (N/2+1 = 2) can still be reached if one sentinel is
lost. See [understanding sentinels][sentinel-tutorial].

- `redis-sentinel-master` / `redis-sentinel-replica` —
  [replication][redis-replication] pair; the replica is the standby that
  sentinels can promote on failover.
- `redis-sentinel-1/2/3` — three sentinel processes monitoring `mymaster`,
  quorum 2; mirrors the 1+1+3 topology running in production.
- `redis-sentinel` stack — aggregator that starts the full cluster in one
  command; waits for all nodes to be healthy before the stack is
  considered up.
- `core-sentinel` — `core` variant whose env points at the sentinel cluster
  (`Redis_SENTINELS`, `Redis_NAME`) instead of the standalone `redis-pwd`;
  keeps `hostname: core` so nginx routes transparently without a new vhost.

[ioredis-sentinel]: https://github.com/redis/ioredis#sentinel
[redis-sentinel]: https://redis.io/docs/latest/operate/oss_and_stack/management/sentinel/
[sentinel-tutorial]: https://redis.io/tutorials/operate/redis-at-scale/high-availability/understanding-sentinels/
[redis-replication]: https://redis.io/docs/latest/operate/oss_and_stack/management/replication/

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -198,7 +198,7 @@ jobs:
 
       - name: Execute e2e tests
         run: |
-          yarn test:low:chrome-desktop -c video=false --env TAGS="${{ matrix.tags }}"
+          yarn test:low:chrome-desktop -c video=false --env TAGS="${{ matrix.tags }}" --env LOG_CONTAINER_NAME="${{ matrix.log_container || 'pc-core-1' }}"
           yarn report
 
       - name: Create artifacts-logs directory
@@ -230,3 +230,6 @@ jobs:
             tags: 'not @hybridge and not @ignore'
           - stack: hybridge
             tags: '@hybridge'
+          - stack: redis-sentinel
+            tags: 'not @hybridge and not @ignore'
+            log_container: pc-core-sentinel-1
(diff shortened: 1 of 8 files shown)


jonathanperret:
J'ai la même remarque et proposition que là : https://github.com/proconnect-gouv/federation/pull/1263#issuecomment-4732720222

PR #1263, opened 2026-06-17
🐳 (docker): add rabbitmq-cluster 

douglasduteil:
**Problem**

Local dev has only a plain single-node `broker` (port 5672). 
Production runs a 3-node RabbitMQ cluster for HA, but there is no way to exercise the `amqp-connection-manager` URL failover path locally.

**Proposal**

`dks switch rabbitmq-cluster` — local replica of the production HA topology so the broker code path (`@hybridge` e2e tests) can be exercised against a clustered broker without deploying to k8s.

- `broker-1` (seed) / `broker-2` / `broker-3` — `rabbitmq:3.12.4-management` nodes sharing `RABBITMQ_ERLANG_COOKIE`, loading the existing `definitions.json` via `rabbitmq-cluster.config`. `broker-2/3` use static peer discovery (classic_config backend) to join `broker-1` automatically on fresh start; the cluster config is isolated to `rabbitmq-cluster.config` so the standalone `broker` service is unaffected.
- `core-cluster` — `core` variant whose env points at the cluster URL list  (`HyyyperbridgeBroker_URLS`); keeps `hostname: core` so nginx routes transparently without a new vhost. `env_file` is a full copy of `core.env` with only the broker URLs swapped (same shape as `core-sentinel.env`).
- `hybridge-rie-cluster` — `hybridge-rie` variant pointing `RieBroker_URLS` at the cluster; joined to both the `rie` and `data` networks so it can resolve the broker hostnames via docker DNS (the plain `hybridge-rie` reaches broker via the rie network gateway + host port forwarding, which doesn't apply to the port-less cluster brokers).
- `rabbitmq-cluster` stack — aggregator that starts the full cluster + cluster-aware services in one command; waits for all broker nodes + the cluster services to be healthy before the stack is considered up.
- `rie` queue declared as a [quorum queue][quorum-queue] (`queueOptions.arguments["x-queue-type"] = "quorum"`) so it is replicated across the cluster nodes — without this, scaling only gives connection-level HA, not data HA. Existing classic queues must be deleted before the first start or the declaration will fail.

**Verified**

`dks switch rabbitmq-cluster` → 3-node cluster (`cluster_status` shows `broker-1`, `broker-2`, `broker-3`), `rie` queue is `quorum` with all 3 nodes as members, leader failover on broker-1 stop works (leader moves to a surviving node), rejoins cleanly when broker-1 comes back.

[quorum-queue]: https://www.rabbitmq.com/docs/quorum-queues

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -230,3 +230,5 @@ jobs:
             tags: 'not @hybridge and not @ignore'
           - stack: hybridge
             tags: '@hybridge'
+          - stack: rabbitmq-cluster
+            tags: '@hybridge'
--- a/docker/builds/broker/rabbitmq-cluster.config
+++ b/docker/builds/broker/rabbitmq-cluster.config
@@ -1,0 +1,14 @@
+[
+  {rabbit, [
+    {loopback_users, []}
+  ]},
+  {autocluster, [
+    {peer_discovery_backend, rabbit_peer_discovery_classic_config}
+  ]},
+  {rabbit_peer_discovery_classic_config, [
+    {nodes, ['rabbit@broker-1', 'rabbit@broker-2', 'rabbit@broker-3']}
+  ]},
+  {rabbitmq_management, [
+    {load_definitions, "/etc/rabbitmq/definitions.json"}
+  ]}
+].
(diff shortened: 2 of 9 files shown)


jonathanperret:
Je trouve que ça fait quand même vraiment beaucoup de duplication à se trimballer dans le repo (et à maintenir…) pour tester un mode de fonctionnement qui devrait être complètement transparent dans 99.9% des cas.

Proposition alternative : laisser la stack de dev/CI comme elle est (broker mono-nœud) et compter sur les tests en stack complète dans `k8s-proconnect` pour lever une éventuelle incompatibilité avec le mode multi-nœud ?

douglasduteil:
We are closing this POC

PR #1264, opened 2026-06-17
✨ (pcf, csmr-rie): add quorum queue type for rabbitmq-cluster

douglasduteil:
**Problem**

RabbitMQ cluster nodes use classic queues by default, which do not replicate data across nodes — a single-node failure loses messages.

**Proposal**

Set `x-queue-type: quorum` on `HyyyperbridgeBroker` and `RieBroker` queues so data is replicated across the cluster and survives node restarts.

### DIFF
--- a/back/apps/core-fca-low/src/config/hyyyperbridge-broker.ts
+++ b/back/apps/core-fca-low/src/config/hyyyperbridge-broker.ts
@@ -6,7 +6,7 @@ const env = new ConfigParser(process.env, "HyyyperbridgeBroker");
 export default {
   payloadEncoding: "base64",
   queue: env.string("QUEUE"),
-  queueOptions: { durable: true },
+  queueOptions: { durable: true, arguments: { "x-queue-type": "quorum" } },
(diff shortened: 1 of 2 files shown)


gdarquie:
Are we sure quorum queues are the right choice here?

I was not familiar with them before reading the documentation. They are intended for long-lived, highly available, and critical queues where message loss would significantly impact system correctness (e.g. orders or voting systems).

In our case, message loss would not have a strong impact on the system, since the user can simply retry the operation. So quorum queues may not be necessary here.

Cf https://www.rabbitmq.com/docs/quorum-queues#use-cases

> The intended use of quorum queues is in topologies where queues typically exist for a long time, need high availability and are critical to certain aspects of an application's architecture.
> Examples of good use cases would be incoming orders in a sales system or votes cast in an electoral system where potentially losing messages would have a significant impact on overall system correctness and function.
> Stock tickers, instant messaging systems and RPC reply queues benefit less or not at all from use of quorum queues.

douglasduteil:
We are closing it for now.

PR #1265, opened 2026-06-17
✨ (pcf): add sentinel support to redis config

douglasduteil:
**Problem**

`redis.ts` reads direct-connect env vars (`HOST`, `PORT`, `PASSWORD`,
`DB`, `TLS`) but not the sentinel ones. The `RedisConfig` DTO already
declares the sentinel fields (`sentinels`, `name`,
`sentinelPassword`) — they just had no env-var source.

**Proposal**

Read `Redis_SENTINELS`, `Redis_NAME`, `Redis_SENTINEL_PASSWORD`. When
`Redis_SENTINELS` is non-empty, populate the sentinel fields and
leave `host`/`port` undefined; otherwise keep the direct-connect
fields.

### DIFF
--- a/back/_doc/env-vars.md
+++ b/back/_doc/env-vars.md
@@ -59,8 +59,11 @@
 | Redis_DB                                        | number        |
 | Redis_ENABLE_TLS_FOR_SENTINEL_MODE              | boolean       |
 | Redis_HOST                                      | string        |
+| Redis_NAME                                      | string        |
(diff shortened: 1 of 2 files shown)


gdarquie (approved):

gdarquie:
Do we need to add some of this env var in k8s-proconnect as well? If yes, we could pair it with the other PR here.

douglasduteil:
This will be done !

PR #1267, opened 2026-06-18
🚚 docker: surface shared compose to dependabot

douglasduteil:
**Problem**

Dependabot is configured to track Docker Compose files at `/`, but no compose file lives at the repo root. The shared stack sits at `docker/compose/shared/shared.yml` — outside dependabot's scope — so it never opens PRs to bump shared service images (postgres, mongo, …). Those versions silently fall behind upstream.

**Proposal**

Rename `docker/compose/shared/shared.yml` to `compose.yaml` so it follows the modern Docker Compose convention. Point dependabot's `docker-compose` directory at `/docker/compose/shared` so it actually finds the shared compose file and starts proposing image updates.


rdubigny (approved):
Suggestion: add a linter configuration for YAML files if you do not want to be the linter.

gdarquie (approved):

PR #1268, opened 2026-06-18
🧱 docker: add base-mongo to shared compose

douglasduteil:
**Problem**

mongo-fca-low is built from a custom Dockerfile that hardcodes --tlsMode requireTLS and --keyFile. Every consumer must carry TLS certs and options just to connect in local dev.

**Proposal**

Add base-mongo to shared.yml: plain mongo:8.2.11, --replSet rs0 only, healthcheck inlined. This is the target state; mongo-fca-low still builds from the Dockerfile until the keyFile setup is solved.


douglasduteil:
Merged in #1283

PR #1282, opened 2026-06-18
🔧 core-fca-low: disable mongo TLS in local dev

douglasduteil:
**Problem**

core-fca-low and the migration runner connect to mongo over TLS in local dev, requiring SSL certs to be mounted and configured.

**Proposal**

Remove Mongoose_TLS* env vars from core.env and core-rie.env so the driver connects without TLS. Mark TLS_CA_FILE as optional in the mongoose config to prevent a startup crash when the var is absent.


douglasduteil:
Merged in #1283

PR #1288, opened 2026-06-19
♻️  Reorganize OIDC provider service

rdubigny:
**Problem**

The `oidc-provider` configuration is split across three files, which makes it difficult to update.

In particular, the separation between `oidc-provider-config.service` and `oidc-provider-config-app.service` no longer makes sense.

Additionally, `oidc-provider-config.dto.ts` redefines all the types from the `oidc-provider` `Configuration` type, with several minor differences.  Type-checking should be sufficient to validate the configuration object, making runtime DTO validation unnecessary.

Finally, `config/oidc-provider.ts` contains a large amount of internal `oidc-provider` configuration that is neither environment-dependent nor shared.  Keeping it available at the application level is therefore unnecessary.

This configuration is also not type-checked correctly, which has allowed unsupported or incorrectly defined options such as `timeout`, `grant_types_supported`, `acceptQueryParamAccessTokens`, and `revocationEndpointAuthMethods`.

**Proposal**

Merge `oidc-provider-config-app.service` and the relevant parts of `config/oidc-provider.ts` into `oidc-provider-config.service` so that all configuration is kept in one place. The file can be split again later using a more logical structure.

Remove the custom `Configuration` type definitions and use the original type provided by `oidc-provider`.

Also fix the usage of the `timeout` option in the `oidc-provider` configuration.


douglasduteil (approved):
Loving it !

PR #1291, opened 2026-06-22
🔧 Fix typing in configuration files

rdubigny:
**Problem**

In configuration files, types are overridden using type assertions, even though we specifically need TypeScript to validate them.

This hides type errors instead of reporting them.

**Proposal**

Use explicitly typed constants and remove type assertions from each configuration file.

Fix the resulting type errors.

**Boy scout**

Replace lifetime environment variable values with a more readable format.


douglasduteil (approved):
Le serial cleaner 🕶️

PR #1302, opened 2026-06-23
🐛 fix: handle domains without MX records according to RFC behavior

BenoitSerrano:
**Problem**
Some domains do not have MX records and are incorrectly considered invalid, even though the RFC specifies that in such cases the domain should be treated as if its MX record points to the domain itself.

**Proposal**
Use a DNS lookup fallback to validate the domain when no MX records are present.

### DIFF
--- a/back/libs/email-validator/src/services/email-validator.service.ts
+++ b/back/libs/email-validator/src/services/email-validator.service.ts
@@ -96,14 +96,30 @@ export class EmailValidatorService {
 
     if (!featureMxResolutionValidation) return false;
 
+    const isDomainReachable = await this.computeIsDomainReachable(emailDomain);
+    return isDomainReachable;
+  }
+
+  private async computeIsDomainReachable(domain: string): Promise<boolean> {
     try {
-      const response = await fetch(
-        `https://dns.google/resolve?name=${encodeURIComponent(emailDomain)}&type=MX`,
+      const mxResponse = await fetch(
+        `https://dns.google/resolve?name=${encodeURIComponent(domain)}&type=MX`,
       );
-      const { Answer } = (await response.json()) as {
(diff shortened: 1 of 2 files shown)


BenoitSerrano:
[Adresse mail invalide](https://trello.com/c/ufFCX6iT/1440-adresse-mail-invalide)

douglasduteil (changes requested):
Julien D. nous a déjà dit non :(
En prod, on peut pas faire d'appel MX, d'où le passage par une couche HTTP pour le faire

back/libs/email-validator/src/services/email-validator.service.ts
```diff
@@ -96,16 +97,23 @@ export class EmailValidatorService {
...
 
+    const isDomainReachable = await this.computeIsDomainReachable(emailDomain);
+    return isDomainReachable;
+  }
+
+  async computeIsDomainReachable(domain: string): Promise<boolean> {
     try {
-      const response = await fetch(
-        `https://dns.google/resolve?name=${encodeURIComponent(emailDomain)}&type=MX`,
-      );
```
rdubigny, line 102: no

douglasduteil (approved):
Better


PS: Not a fan of forced private method access

```ts
const result = await service["computeIsDomainReachable"](
  "does-not-exist-1234567890.fr",
);
```

BenoitSerrano:
> Better
> 
> PS: Not a fan of forced private method access
> 
> ```ts
> const result = await service["computeIsDomainReachable"](
>   "does-not-exist-1234567890.fr",
> );
> ```

Yeah me neither, but the solution would be either :
- make it public (not a fan)
- do not test the function specifically, only test it via the higher ranked functions that use it (harder to test, less readable)

Unless you see something else?

PR #1306, opened 2026-06-24
↗️ update to Mongo 8.3.4

douglasduteil:
**Problem**

- We are not using the production mongo image.
- dependabot is stuck unable to update mongo image tag.

**Proposal**

- The production secure bitnami/mongo image.
- Pin version with a latest tag comment to hind dependabot from updating the pin with the future latest version.
- `docker run --rm bitnami/mongodb:latest mongod --version` to check the version of the mongo in the image.
- `docker pull bitnami/mongodb:latest | grep Digest` to determine the digest of the latest mongo image.


douglasduteil:
Out of date for now

PR #1310, opened 2026-06-25
✨ feat: add collaborator management for partner users

BenoitSerrano:
**Problem**
Users of the Partner Portal cannot invite collaborators to manage their Service Providers.

**Proposal**
Add a `collaborators` field to the database and provide the ability to manage it through the admin interface.

### DIFF
--- a/admin/src/service-provider/dto/service-provider-input.dto.spec.ts
+++ b/admin/src/service-provider/dto/service-provider-input.dto.spec.ts
@@ -12,6 +12,7 @@ describe("Service Provider Input (Data Transfer Object)", () => {
     active: "true",
     type: "private",
     scopes: ["given_name"],
+    collaborators: [],
   };
 
   it("should validate if all properties are correct", async () => {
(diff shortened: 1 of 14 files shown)


BenoitSerrano:
[Ajouter un champ "collaborators" éditable dans l'admin](https://trello.com/c/HqtuPWYi/1463-ajouter-un-champ-collaborators-%C3%A9ditable-dans-ladmin)

rdubigny (approved):

admin/src/service-provider/dto/service-provider-input.dto.spec.ts
```diff
@@ -12,6 +12,7 @@ describe("Service Provider Input (Data Transfer Object)", () => {
     active: "true",
     type: "private",
     scopes: ["given_name"],
+    collaborators: [],
```
rdubigny, line 15: Thoughts: I would have liked to see more tests for this field validation, such as collaborators with empty string, `null`, or `undefined` values.

PR #1311, opened 2026-06-26
⬇️ chore: align local PostgreSQL version with production

BenoitSerrano:
**Problem**
The local PostgreSQL version (18) is newer than the production version (16), which may lead to inconsistencies between development and production environments.

**Proposal**
Downgrade the local PostgreSQL version to 16 so it matches the production environment.


rdubigny (approved):
todo: you should also pin this version in dependabot so it does not suggest other irrelevant update.

cc @douglasduteil

rdubigny (approved):

PR #1312, opened 2026-06-26
✨ feat: support collaborator management for OIDC clients

BenoitSerrano:
**Problem**
The Espace Partenaires cannot add collaborators to OIDC clients.

**Proposal**
Allow editing the `collaborators` field and grant access to the specified email addresses.

### DIFF
--- a/pcdbapi/test_api.py
+++ b/pcdbapi/test_api.py
@@ -9,10 +9,8 @@ import time
 
 import pytest
 from asgi_lifespan import LifespanManager
-from fastapi import Request
 from httpx import ASGITransport, AsyncClient, Response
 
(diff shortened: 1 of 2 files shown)


pcdbapi/test_api.py
```diff
@@ -9,10 +9,8 @@
 
 import pytest
 from asgi_lifespan import LifespanManager
-from fastapi import Request
```
gdarquie, line 12: question: wow, there were useless import previous to your changes?

gdarquie (approved):

pcdbapi/test_api.py
BenoitSerrano: I know right, shocking

PR #1335, opened 2026-06-30
✨ feat: add MFA compliance flag for Identity Providers

BenoitSerrano:
**Problem**
Identity Providers that are not MFA-compliant fail authentication when a Service Provider requires MFA.

**Proposal**
Add an `isMfaCompliant` field to Identity Providers so MFA requirements can be handled appropriately.


BenoitSerrano:
[ETQU, ajouter un champ mfa_compliant sur les FI](https://trello.com/c/xUvLYeeO/1473-etqu-ajouter-un-champ-mfacompliant-sur-les-fi)

PR #1338, opened 2026-07-04
🐛 fix: skip MFA ACR requests for non-compliant Identity Providers

BenoitSerrano:
**Problem**
Identity Providers that are not MFA-compliant cannot handle requested ACRs related to MFA, causing authentication failures.

**Proposal**
For Identity Providers flagged as non-MFA-compliant, do not send MFA-related requested ACRs during authentication.


BenoitSerrano:
Oupsie the tests fail
I'm not sure it has a link with the PR, it should be more about changes in the fixtures!

PR #1342, opened 2026-07-06
🏷️ back: add truthful type overloads to ConfigParser methods

douglasduteil:
**Problem**

TS6 strict mode flags lying return types in ConfigParser: string() could return undefined but declared `string`, callers masked it with `as string`/`as boolean`/`as number` casts that hid real type unsafety across all config files.

**Proposal**

Add proper overloads so single-arg calls (`string(path)`, `boolean(path)`, `number(path)`) return the truthful required type and throw on missing/invalid values. Two-arg calls with `undefinedIfEmpty: true` return `string | undefined` for optional config values. Remove lying casts from all core-fca-low config files.


back/apps/core-fca-low/src/config/app.ts
```diff
@@ -8,13 +8,13 @@ const appConfig: AppConfig = {
...
   assetsDsfrPaths: env.json("DSFR_ASSETS_PATHS"),
-  assetsCacheTtl: env.number("ASSETS_CACHE_TTL"),
+  assetsCacheTtl: env.number("ASSETS_CACHE_TTL", true),
   viewsPaths: env.json("VIEWS_PATHS"),
   httpsOptions: {
     key: env.file("HTTPS_SERVER_KEY", true),
     cert: env.file("HTTPS_SERVER_CERT", true),
   },
-  fqdn: process.env.FQDN,
+  fqdn: process.env.FQDN as string,
```
BenoitSerrano, line 17: Is this casting true? Don't we want a `?? ""` here?

back/libs/config/src/helpers/config-parser.ts
```diff
@@ -31,30 +40,36 @@ export class ConfigParser {
...
   }
 
   stringArray(path: string): string[] {
     const parsedString = this.string(path, true);
     return parsedString?.split(",") || [];
   }
 
-  number(path: string, undefinedIfEmpty: boolean = false): number {
+  number(path: string): number;
+  number(path: string, undefinedIfEmpty: true): number | undefined;
```
BenoitSerrano, line 61: Not a fan of not-named function parameters when their utility is not obvious. For instance, here I would expect a boolean second parameter to be a *default value*, I would never infer its utility from the signature only. 
I would do something like `number(path: string, options?: { undefinedIfEmpty: true}): number | undefined;`

back/apps/core-fca-low/src/config/app.ts
douglasduteil: The value is somethinkg like `FQDN=core-fca-low.docker.dev-franceconnect.fr`
I guess that it actually should be a `env.string("FQDN")` 
An additionnal migration can be `FQDN` => `App_FQDN`

back/libs/config/src/helpers/config-parser.ts
douglasduteil: I agree. this PR manly try to give a TS6 compatibility on existing :hankey:.
I expect that we chose to go on and unify our approach later relying on Zod `z.number().optional()` obvious api.

douglasduteil: https://github.com/nestjs/nest 12 (wip in 16391 pull request) will help but I don't know if we will wait for it

douglasduteil:
[Continuer la migration TS5 -> 6 -> 7 sur Fédération](https://trello.com/c/4utJWf1D/1570-continuer-la-migration-ts5-6-7-sur-f%C3%A9d%C3%A9ration)

PR #1344, opened 2026-07-06
🗃️ chore: backfill app owners as collaborators

BenoitSerrano:
**Problem**
The application owner's email should be included by default in the `collaborators` field, but existing applications do not have it.

**Proposal**
Add a migration to populate the `collaborators` field with the application owner's email for existing records.

### DIFF
--- a/back/migrations/20260706134953-move_client.email_to_collaborators.ts
+++ b/back/migrations/20260706134953-move_client.email_to_collaborators.ts
@@ -1,0 +1,15 @@
+import type { Db } from "mongodb";
+
+export const up = async (db: Db) => {
+  const clients = await db.collection("client").find({}).toArray();
+  for (const client of clients) {
+    const emails = client.email?.split("\n").filter(Boolean) || [];
+    await db
+      .collection("client")
+      .updateOne({ _id: client._id }, { $set: { collaborators: emails } });
+  }
+};
+
+export const down = async (db: Db) => {
+  await db.collection("client").updateMany({}, { $set: { collaborators: [] } });
+};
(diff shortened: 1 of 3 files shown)


back/migrations/20260706134953-move_client.email_to_collaborators.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * @param db {import('mongodb').Db}
+ * @param client {import('mongodb').MongoClient}
+ * @returns {Promise<void>}
+ */
+export const up = async (db, client) => {
+  // TODO write your migration here.
+  // See https://github.com/seppevs/migrate-mongo/#creating-a-new-migration-script
+  // Example:
```
douglasduteil:
```suggestion
```

back/migrations/20260706134953-move_client.email_to_collaborators.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * @param db {import('mongodb').Db}
+ * @param client {import('mongodb').MongoClient}
+ * @returns {Promise<void>}
+ */
```
douglasduteil: I think we should prefer direct import type as we are in a ts file 
Example in the first migration file
```
import type { IdentityProvider } from "@fc/identity-provider-adapter-mongo/schemas";
import type { Db } from "mongodb";
```

douglasduteil (changes requested):

back/migrations/20260706134953-move_client.email_to_collaborators.ts
```diff
@@ -0,0 +1,15 @@
+import type { Db } from "mongodb";
+
+export const up = async (db: Db) => {
+  const clients = await db.collection("client").find({}).toArray();
+  for (const client of clients) {
+    const emails = client.email?.split("\n").filter(Boolean) || [];
```
douglasduteil, line 6: Is this typed ? Don't you need `db.collection<...>("client").` ?

douglasduteil (approved):
Look better

PR #1362, opened 2026-07-07
🔒 security: remove unsafe-inline from Content Security Policy

BenoitSerrano:
**Problem**
The current Content Security Policy includes the `unsafe-inline` directive, which weakens the application's security by allowing inline scripts.

**Proposal**
Remove the `unsafe-inline` directive from the Content Security Policy.


BenoitSerrano:
[Supprimer les CSP unsafe-inline pour les scripts et style](https://trello.com/c/6ixYXbHk/1472-supprimer-les-csp-unsafe-inline-pour-les-scripts-et-style)

douglasduteil (approved):
LGTM

PR #1364, opened 2026-07-08
🔧 Remove non essential @k8s tags

gdarquie:
We only start minimal tests in k8s-proconnect CI, so there's no need to keep these tags on all test scenarios. 

This cleanup removes unnecessary "k8s" tags from quality test files to reduce clutter and maintain only the essential markers for CI orchestration.


gdarquie:
[Tests K8S : réduire le volume de tests](https://trello.com/c/aoDkHor2/1490-tests-k8s-r%C3%A9duire-le-volume-de-tests)

PR #1367, opened 2026-07-10
🦺 pcdbapi: reject empty collaborators list

douglasduteil:
**Problem**

PATCH accepts collaborators: [] even though collaborators is the only access path since dcb796b0e. Emptying it permanently locks everyone out of the app, with no recovery short of manual DB surgery.

**Proposal**

Require at least one entry in collaborators through the pydantic model, answering 422 on an empty list.


BenoitSerrano (approved):
Great job! :D

PR #1371, opened 2026-07-13
🚸 quiet the session-not-found error page

douglasduteil:
**Problem**

The oidc-provider-error:session-not-found page shows an error code, id and message for a failure we cannot help with (the interaction id is simply gone from redis), generating noise and user reports.

**Proposal**

Render only the friendly description with a "Revenir à l'étape précédente" (history.back) button; keep the technical details in the logs. Add generic "je vois" / "je ne vois pas" cypress steps and use them in the affected scenarios.

--- 

image

### DIFF
--- a/back/apps/core-fca-low/src/views/error.ejs
+++ b/back/apps/core-fca-low/src/views/error.ejs
@@ -76,7 +76,7 @@
               <% } %>
             </div>
 
-            <% if (locals.exceptionDisplay.displayContact || locals.interactionErrorUrl) { %>
+            <% if (locals.exceptionDisplay.displayContact || locals.interactionErrorUrl || locals.exceptionDisplay.mainAction === 'goBack') { %>
               <section>
                 <div class="fr-mt-4w">
                   <% if (locals.exceptionDisplay.displayContact) { %>
@@ -91,6 +91,11 @@
                     <% if (locals.exceptionDisplay.displayContact) { %>
                       <a class="fr-btn<%= locals.exceptionDisplay.mainAction === 'contact' ? '' : ' fr-btn--tertiary' %> fr-icon-flag-line fr-btn--icon-right fr-mr-1w" href="<%= locals.exceptionDisplay.contactHref %>">
                         Signaler l'erreur
+                      </a>
+                    <% } %>
+                    <% if (!locals.interactionErrorUrl && locals.exceptionDisplay.mainAction === 'goBack') { %>
+                      <a class="fr-btn" href="javascript:history.back()" data-testid="go-back-button">
+                        Revenir à l'étape précédente
                       </a>
                     <% } %>
                     <% if (locals.interactionErrorUrl) { %>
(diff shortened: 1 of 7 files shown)


douglasduteil:
[Erreur oidc-provider-error:session-not-found](https://trello.com/c/qp3ts8FK/1461-erreur-oidc-provider-errorsession-not-found)

Anais-Tailhade (approved):
it's ok for me. Thank's

back/libs/oidc-provider/src/filters/oidc-provider-session-not-found-exception.filter.spec.ts
```diff
@@ -95,5 +95,21 @@ describe("OidcProviderSessionNotFoundExceptionFilter", () => {
       // Then
       expect(resMock.render).toHaveBeenCalledOnce();
     });
+
```
BenoitSerrano, line 98: I don't understand the purpose of adding a new test (as opposed to adding an `expect` statement to the previous one)

quality/cypress/support/common/steps/visibility-steps.ts
```diff
@@ -0,0 +1,9 @@
+import { Then } from "@badeball/cypress-cucumber-preprocessor";
+
+Then("je vois {string}", function (text: string) {
+  cy.contains(text).should("be.visible");
+});
+
+Then("je ne vois pas {string}", function (text: string) {
+  cy.contains(text).should("not.exist");
+});
```
rdubigny, line 9: Suggestion: while having generic steps sounds like a good idea, it is not OK to introduce new patterns only for your use case.

If your intention is to replace the existing specific accessors with generic ones, this is not the right way to do it. We will most likely end up with two patterns being used arbitrarily.

If your intention is to add a new way to access things, you are adding too much code for humans to read. Please consider factoring it.

back/apps/core-fca-low/src/views/error.ejs
```diff
@@ -93,6 +93,11 @@
                         Signaler l'erreur
                       </a>
                     <% } %>
+                    <% if (!locals.interactionErrorUrl && locals.exceptionDisplay.mainAction === 'goBack') { %>
+                      <a class="fr-btn" href="javascript:history.back()" data-testid="go-back-button">
+                        Revenir à l'étape précédente
```
rdubigny, line 98: issue: the error description suggest to close the window, the call to action should help the user do that instead of proposing another action.

back/libs/oidc-provider/src/filters/oidc-provider-session-not-found-exception.filter.ts
```diff
@@ -22,15 +22,12 @@ export class OidcProviderSessionNotFoundExceptionFilter extends BaseExceptionFil
     const statusCode = 400;
 
     const errorPageParams: ErrorPageParams = {
-      error: {
-        code,
-        id,
-        message,
-      },
+      error: {},
```
rdubigny, line 25: Suggestion: this looks like a "poussière sous le tapis" technique. I suggest changing the log level of the associated log instead of removing information that could help with debugging.

rdubigny (changes requested):
There is an underlying problem with this proposal: the problem you are trying to solve is incomplete.

If you refer to the original Crisp error, the user states that they are experiencing this error **repeatedly**, suggesting that there is an underlying issue with the way `oidc-provider` manages its sessions.

It is likely that adding a “go back” button will not solve the original problem.

douglasduteil:
OK I'm closing this proposal

PR #1372, opened 2026-07-13
🐛 fix: disable autofill on collaborators field

BenoitSerrano:
**Problem**
Some versions of Bitwarden automatically fill the `collaborators` field with the user's email address when autofilling the TOTP field.

**Proposal**
Disable autocomplete on the `collaborators` field to prevent incorrect autofill behavior.

### DIFF
--- a/admin/views/service-provider/creation.ejs
+++ b/admin/views/service-provider/creation.ejs
@@ -117,7 +117,18 @@
             <section class="mb-3 row">
               <label class="col-2 col-form-label" for="collaborators">Collaborateurices (une adresse e-mail par ligne)</label>
               <div class="col-10">
-                <textarea id="collaborators" class="form-control <%= locals.messages.errors && messages.errors[0].collaborators && 'is-invalid' %>" pattern="<%= VALID_EMAILS_REGEX %>" placeholder="Adresses e-mail des collaborateurices" name="collaborators"><%= locals.messages.values && messages.values[0].collaborators %></textarea>
+                <textarea
+                  id="collaborators"
+                  class="form-control
+                  <%= locals.messages.errors && messages.errors[0].collaborators && 'is-invalid' %>"
+                  pattern="<%= VALID_EMAILS_REGEX %>"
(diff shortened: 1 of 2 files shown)


BenoitSerrano:
[supprimer l'autocomplete de l'email sur le nouveau champ collaborator](https://trello.com/c/Lj2a9Ac4/1504-supprimer-lautocomplete-de-lemail-sur-le-nouveau-champ-collaborator)

admin/views/service-provider/creation.ejs
```diff
@@ -117,7 +117,7 @@
             <section class="mb-3 row">
               <label class="col-2 col-form-label" for="collaborators">Collaborateurices (une adresse e-mail par ligne)</label>
               <div class="col-10">
-                <textarea id="collaborators" class="form-control <%= locals.messages.errors && messages.errors[0].collaborators && 'is-invalid' %>" pattern="<%= VALID_EMAILS_REGEX %>" placeholder="Adresses e-mail des collaborateurices" name="collaborators"><%= locals.messages.values && messages.values[0].collaborators %></textarea>
+                <textarea id="collaborators" class="form-control <%= locals.messages.errors && messages.errors[0].collaborators && 'is-invalid' %>" pattern="<%= VALID_EMAILS_REGEX %>" placeholder="Adresses e-mail des collaborateurices" name="collaborators" autocomplete="off"><%= locals.messages.values && messages.values[0].collaborators %></textarea>
```
rdubigny: question: email field also include these attributes:

```html
    data-1p-ignore="true"
    data-bwignore="true"
    data-form-type="other"
```

Why are you not including them here?

Have you tested it with 1password that I use? (If not I can test it for you)

BenoitSerrano: I was not including these because I didn't know what they did hehe
I now do and they specically tell 1-password and Bitwarden not to consider this input, and most browsers *not* to consider this field as a known category.
I don't have 1password :x could you please try it on your browser? :D

rdubigny: it works!

PR #1391, opened 2026-07-21
🩺 Add livez/readyz healthcheck to admin

douglasduteil:
**Problem**

admin has no /livez or /readyz endpoints, unlike core-fca-low and csmr-rie, so Kubernetes has no way to probe its readiness before routing traffic to it.

**Proposal**

Add a HealthController modeled on the existing core-fca-low/csmr-rie pattern: /livez always returns ok, /readyz checks both TypeORM connections (postgresDatabase and mongoDatabase) and returns 503 with a verbose per-check breakdown when either is not initialized.

### DIFF
--- a/admin/src/health.controller.spec.ts
+++ b/admin/src/health.controller.spec.ts
@@ -1,0 +1,145 @@
+import { HttpStatus } from "@nestjs/common";
+import { Test, TestingModule } from "@nestjs/testing";
+import { getDataSourceToken } from "@nestjs/typeorm";
+import { HealthController } from "./health.controller";
+
+describe("HealthController", () => {
+  let controller: HealthController;
+
+  const databaseMock = {
+    isInitialized: true,
+    query: jest.fn(),
+  };
+  const mongoRepositoryMock = {
+    count: jest.fn(),
+  };
+  const mongoDatabaseMock = {
+    isInitialized: true,
+    manager: {
+      getMongoRepository: jest.fn(() => mongoRepositoryMock),
+    },
+  };
+
+  const resMock = {
+    status: jest.fn().mockReturnThis(),
+  };
+
+  beforeEach(async () => {
+    jest.resetAllMocks();
+    databaseMock.isInitialized = true;
+    databaseMock.query.mockResolvedValue([{ "?column?": 1 }]);
+    mongoDatabaseMock.isInitialized = true;
+    mongoDatabaseMock.manager.getMongoRepository.mockReturnValue(
+      mongoRepositoryMock,
+    );
+    mongoRepositoryMock.count.mockResolvedValue(0);
+
+    const app: TestingModule = await Test.createTestingModule({
+      controllers: [HealthController],
+      providers: [
+        { provide: getDataSourceToken(), useValue: databaseMock },
+        {
+          provide: getDataSourceToken("fc-mongo"),
+          useValue: mongoDatabaseMock,
+        },
+      ],
+    }).compile();
+
+    controller = app.get<HealthController>(HealthController);
+  });
+
+  describe("livez()", () => {
+    it("should return 'ok'", () => {
+      expect(controller.livez()).toBe("ok");
+    });
+  });
+
+  describe("readyz()", () => {
+    it("should return 'ok' when both connections are initialized and reachable", async () => {
+      const result = await controller.readyz(resMock as any);
+      expect(databaseMock.query).toHaveBeenCalledWith("SELECT 1");
+      expect(mongoRepositoryMock.count).toHaveBeenCalled();
+      expect(resMock.status).not.toHaveBeenCalled();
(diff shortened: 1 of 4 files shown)


gdarquie:
[Ajouter une route readiness pour l'admin](https://trello.com/c/Nl2rPvVx/1512-ajouter-une-route-readiness-pour-ladmin)

admin/src/health.controller.ts
```diff
@@ -0,0 +1,81 @@
...
+      await this.database.query("SELECT 1");
+    },
+    "fc-mongo": async () => {
+      if (!this.mongoDatabase.isInitialized) {
+        throw new Error("fc-mongo connection not initialized");
+      }
+      await this.mongoDatabase.manager
+        .getMongoRepository(Configuration)
+        .count();
+    },
```
gdarquie, line 37: we also need to check psql connection, isn't it?

gdarquie (commented):
Have you seen this commit, @douglasduteil? https://github.com/proconnect-gouv/federation/compare/main...add-admin-readiness

I had started implementing this health check when the CI was blocked, but I stopped once we realized the main issue was with the resources.

I used the same logic as your backend implementation, but without the tests or refactoring.

admin/src/health.controller.ts
```diff
@@ -0,0 +1,81 @@
...
+} from "@nestjs/common";
+import { InjectDataSource } from "@nestjs/typeorm";
+import { type Response } from "express";
+import { DataSource } from "typeorm";
+import { Configuration } from "./configuration/entity/configuration.mongodb.entity";
+
+@Controller()
+export class HealthController {
+  constructor(
+    @InjectDataSource() private readonly database: DataSource,
```
gdarquie:
```suggestion
    @InjectDataSource() private readonly postgresDatabase: DataSource,
```

admin/src/health.controller.ts
gdarquie: Ok la connexion par défaut est postgres.

admin/src/health.controller.ts
```diff
@@ -0,0 +1,81 @@
...
+    @InjectDataSource("fc-mongo") private readonly mongoDatabase: DataSource,
+  ) {}
+
+  checks = {
+    database: async () => {
+      if (!this.database.isInitialized) {
+        throw new Error("database connection not initialized");
```
gdarquie:
```suggestion
        throw new Error("PostgreSQL default database connection not initialized");
```

admin/src/health.controller.ts
```diff
@@ -0,0 +1,81 @@
...
+      }
+      await this.database.query("SELECT 1");
+    },
+    "fc-mongo": async () => {
+      if (!this.mongoDatabase.isInitialized) {
+        throw new Error("fc-mongo connection not initialized");
```
gdarquie:
```suggestion
        throw new Error("MongoDB fc-mongo connection not initialized");
```

gdarquie (approved):

admin/src/health.controller.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+          const result = await controller.readyz(resMock as any, "");
+          expect(resMock.status).toHaveBeenCalledWith(
+            HttpStatus.SERVICE_UNAVAILABLE,
+          );
+          expect(result).toBe(expectedMessage);
+        },
+      );
+    });
+  });
+});
```
gdarquie, line 145: question: no e2e cypress tests?

douglasduteil: Nope. Integration test in the futur we hope :)

admin/src/health.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+
+@Controller()
+export class HealthController {
+  constructor(
+    @InjectDataSource() private readonly postgresDatabase: DataSource,
+    @InjectDataSource("fc-mongo") private readonly mongoDatabase: DataSource,
+  ) {}
+
+  checks = {
+    database: async () => {
```
rdubigny: Suggestion: since this is what external users will be able to read, I suggest using an explicit name: `postgresDatabase`.

admin/src/health.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+      if (!this.postgresDatabase.isInitialized) {
+        throw new Error(
+          "PostgreSQL default database connection not initialized",
+        );
+      }
+      await this.postgresDatabase.query("SELECT 1");
+    },
+    "fc-mongo": async () => {
```
rdubigny: Suggestion: since this is what external users will be able to read, I suggest using an explicit name: `mongoDatabase`.

gdarquie (approved):

rdubigny (approved):
If you like seeing me happy, plz squash!

PR #1408, opened 2026-07-21
♻️ Inject identity provider through abstract adapter

douglasduteil:
**Problem**

Consumers hard-import the Mongo adapter, blocking alternative implementations.

**Proposal**

Type consumers against an IdentityProviderAdapter abstract class used as injection token.


back/libs/oidc-client/src/interfaces/identity-provider-adapter.interface.ts
```diff
@@ -1,14 +1,20 @@
 import { IdentityProviderMetadata } from "@fc/oidc";
 
-export interface IIdentityProviderAdapter {
-  getById(
+export abstract class IdentityProviderAdapter {
+  abstract getById(
```
rdubigny, line 4: Suggestion: as this is no longer an interface, maybe you should consider another folder for these files, especially if you plan to generalize this pattern to all adapters.

rdubigny (commented):
Can you clarify in the problem section of the commit body why we need to be ready to support multiple implementations at the same time?

When I read the problem you want to solve:

> Consumers hard-import the Mongo adapter, blocking alternative implementations.

I wonder why we are not simply hard-importing the alternative implementation directly, without requiring an abstraction layer.

I might be missing something. I would welcome any additional explanation.

douglasduteil:
@rdubigny I have a dream that one day this federation code can be imported in ci and locally without the need of mongodb (futur postgres) or redis. 
I can finally remove the JS `http://localhost:3000/___dev___/auth.agentconnect.gouv.fr/api/v2` routes from my ProConnect clients and the direct dependence between a CI and `federation-sandbox.proconnect.gouv.fr`. 
For this dream to become reality, we need to make the buisness logic agnostic to Mongodb.

back/libs/email-validator/src/services/email-validator.service.ts
```diff
@@ -16,15 +16,14 @@ import { EmailValidatorConfig } from "../dto";
 export class EmailValidatorService {
   constructor(
     private readonly logger: LoggerService,
-    private readonly identityProviderAdapterMongoService: IdentityProviderAdapterMongoService,
+    private readonly identityProvider: IdentityProviderAdapter,
```
BenoitSerrano, line 19: I'm not a fan of this naming
this.identityProviderAdapter ?
or
this.identityProviderRepository?
this.identityProviderManager?
this.identityProviderService?

I'm not a fan of the latest two either but I find this.identityProvider confusing as this variable name is used everywhere throughout the app

BenoitSerrano (commented):
I'd be very pleased to help you get this work merged! Postgres is comiiiing

douglasduteil:
Feel free to help on this @BenoitSerrano
I would welcome your commits on this branch :)

PR #1410, opened 2026-07-22
🚚 Extract logout confirm script to external file

douglasduteil:
**Problem**

The logout confirmation page rendered by OidcProviderConfigService inlines a <script> block directly in the HTML template. The app's CSP scriptSrc policy has no 'unsafe-inline' allowance, and every other page-level script in core-fca-low already lives under public/js/ and is loaded via <script src=...>.

**Proposal**

Move the auto-submit logic into public/js/end-session-confirm.js, served as a static asset, and reference it with a src attribute instead of inlining it.

### DIFF
--- a/back/apps/core-fca-low/src/public/js/end-session-confirm.js
+++ b/back/apps/core-fca-low/src/public/js/end-session-confirm.js
@@ -1,0 +1,7 @@
+var form = document.forms[0];
+var input = document.createElement("input");
+input.type = "hidden";
+input.name = "logout";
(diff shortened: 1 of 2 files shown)


rebeccadumazert (approved):

douglasduteil:
[Réparer la déconnexion](https://trello.com/c/y9s6ZSSM/1550-r%C3%A9parer-la-d%C3%A9connexion)

rdubigny (approved):

PR #1412, opened 2026-07-27
🐛 Guard email-to-collaborators migration against clobbering seeded collaborators

douglasduteil:
**Problem**
`20260706134953-move_client.email_to_collaborators.ts` runs `find({})` over every client and unconditionally `$set`s `collaborators`, defaulting to `[]` when `client.email` is absent. On environments where `collaborators` is seeded another way (k8s-proconnect fixtures, #1312 collaborators feature), the migration wipes it back to `[]` the first time it fires — same root cause as the #1344/dcb796b0e incident, resurfacing via fresh-cluster CI on helm-fca-low.

**Proposal**
Scope the migration to `find({ email: { $exists: true } })` and merge into `collaborators` with `$addToSet` instead of overwriting, making it a no-op for documents without `client.email`. `down` is left as-is (blind `$set: { collaborators: [] }`) since a precise inverse isn't feasible with `$addToSet`.

### DIFF
--- a/back/migrations/20260706134953-move_client.email_to_collaborators.ts
+++ b/back/migrations/20260706134953-move_client.email_to_collaborators.ts
@@ -1,12 +1,19 @@
 import type { Db } from "mongodb";
 
 export const up = async (db: Db) => {
-  const clients = await db.collection("client").find({}).toArray();
+  const clients = await db
+    .collection("client")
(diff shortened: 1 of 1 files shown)


BenoitSerrano (approved):

rdubigny (commented):
I would be more comfortable with fewer changes in this PR. The changes on line 4 look sufficient.

That said, it sounds good for the intended usage.

PR #1427, opened 2026-07-29
🔥 Remove pcdbapi service

douglasduteil:
**Problem**

The pcdbapi service (Python FastAPI OIDC client manager) is no longer used. Keeping it adds maintenance burden, CI minutes, and dependency surface for a dead code path.

**Proposal**

Remove the entire pcdbapi/ directory along with its CI workflow, Docker build matrix entry, and Dependabot config entries.


rdubigny (approved):
🎉

PR #1434, opened 2026-07-29
🐛 Improve unexpected error handling

rdubigny:
**Problem**

In the email verification service, all errors are caught in the controller and the email verification page is rendered to display the error.

However, if an unexpected error occurs, `countdownEndDate` may be undefined, causing the template rendering to crash. This results in an incomprehensible error message for users and makes debugging more difficult.

**Proposal**

Use the EJS `locals` variable, which is more resilient when optional values are missing, to avoid rendering errors.

Proper error handling will be implemented in a future PR.


BenoitSerrano (approved):

douglasduteil (approved):
🪭 🆗 🪭

PR #1436, opened 2026-07-29
remove unused baseurl email template parameter

BenoitSerrano:
**Problem**
The `baseurl` parameter was required for every email template, even though it was no longer used.

**Proposal**
Remove the unused `baseurl` parameter from the email template API.


BenoitSerrano:
[Follow up : OTP mail](https://trello.com/c/cAxXObbR/1557-follow-up-otp-mail)

rdubigny (approved):

PR #1447, opened 2026-07-30
🐛 Derive cypress FEDERATION_DIR from config file location

douglasduteil:
**Problem**

FEDERATION_DIR and LOG_FILE_PATH were built from
`${PC_ROOT}/federation`, hardcoding the repo directory name. Any checkout not literally named `federation` (e.g. a git worktree) silently points Cypress at a sibling directory's log file instead of its own, making `hasBusinessLog` assertions fail against a stale or empty log with no indication why.

**Proposal**

Derive REPO_ROOT from cypress.config.js's own location (`import.meta.dirname`) instead of assuming the checkout name, so the log path always matches the repo actually being run.

### DIFF
--- a/admin/cypress.config.js
+++ b/admin/cypress.config.js
@@ -1,5 +1,8 @@
 import { defineConfig } from "cypress";
+import path from "path";
 import pluginConfig from "./cypress/plugins";
+
+const REPOSITORY_ROOT = path.resolve(import.meta.dirname, "..");
 
 export default defineConfig({
   chromeWebSecurity: false,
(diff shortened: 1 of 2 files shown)


douglasduteil:
This is specially useful for git worktree :)

admin/cypress.config.js
```diff
@@ -1,6 +1,9 @@
 import { defineConfig } from "cypress";
+import path from "path";
 import pluginConfig from "./cypress/plugins";
 
+const REPO_ROOT = path.resolve(import.meta.dirname, "..");
```
BenoitSerrano: :suggestion REPOSITORY_ROOT?

BenoitSerrano:
Don't we have the same issue in the `quality` directory?

BenoitSerrano (approved):

rdubigny (approved):

PR #1449, opened 2026-07-31
🐛 Replace admin e2e pg-db clone with TRUNCATE + reseed

douglasduteil:
**Problem**

Cypress e2e resets recreated pg-db from a pg-backup template DB via CREATE DATABASE ... WITH TEMPLATE, which requires zero connections on the source. The scripts only terminated connections on the target, causing an intermittent CI failure ("source database is being accessed by other users") when a leftover session (fixtures loader) was still attached to pg-db.

Terminating connections on the source fixes that race, but it also kills the admin app's own live pooled connections on every single test reset, forcing a silent reconnect each time - a second, structural source of flakiness that no amount of
pg_terminate_backend closes off, since dropping/recreating a database inherently requires zero connections to it.

**Proposal**

Stop cloning the whole database. Capture a data-only pg_dump of the three e2e-relevant tables once fixtures are loaded, then reset by TRUNCATE + replay instead of DROP/CREATE DATABASE. TRUNCATE takes a table-level lock that existing pooled connections simply wait on rather than getting killed, so the app's connections are never disrupted. Mirrors the project's existing Mongo reset convention (wipe in place + replay seed scripts) instead of a whole-DB clone.

Also drops the sleep 10 in up.sh that used to paper over the race - no longer needed since the reset is now deterministic.

### DIFF
--- a/admin/cypress/support/apply-db-backup.sh
+++ b/admin/cypress/support/apply-db-backup.sh
@@ -1,6 +1,4 @@
 #!/bin/bash
 
-psql postgres -U pg-user -c 'REVOKE CONNECT ON DATABASE "pg-db" FROM public'
-psql postgres -U pg-user -c "SELECT \"pid\", pg_terminate_backend(pid) FROM \"pg_stat_activity\" WHERE datname='pg-db' AND pid <> pg_backend_pid();"
-psql postgres -U pg-user -c 'DROP DATABASE  IF EXISTS "pg-db"'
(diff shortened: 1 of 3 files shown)


douglasduteil:
This is fixing the many CI error we have with

```

ERROR:  database "pg-backup" does not exist
 pid | pg_terminate_backend 
-----+----------------------
(0 rows)

NOTICE:  database "pg-backup" does not exist, skipping
DROP DATABASE
ERROR:  source database "pg-db" is being accessed by other users
DETAIL:  There is 1 other session using the database.
```

BenoitSerrano (approved):
Wow, a long-awaited fix!

PR #1450, opened 2026-07-31
🔒 security: reuse MFA sessions when ACR requirements are satisfied

BenoitSerrano:
**Problem**
Users receive a new MFA email every time they authenticate to another Service Provider requiring MFA, even when their existing session already satisfies the requested assurance level.

**Proposal**
Allow session reuse when the ACR values present in the existing session satisfy the requirements of the new Service Provider, except for sensitive ACR values (`eidas2` and `eidas3`) which require a fresh authentication session.

### DIFF
--- a/back/apps/core-fca-low/src/controllers/interaction.controller.ts
+++ b/back/apps/core-fca-low/src/controllers/interaction.controller.ts
@@ -109,8 +109,15 @@ export class InteractionController {
     const isSessionOpenedWithHintedIdp =
       !idpHint || userSession.get("idpId") === hintedIdp.uid;
 
-    const isEssentialAcrSatisfied =
-      this.oidcAcr.isEssentialAcrSatisfied(interaction);
+    const { acrClaims } =
+      this.oidcAcr.getFilteredAcrParamsFromInteraction(interaction);
+    const spEssentialAcr =
+      acrClaims?.value || acrClaims?.values?.join(" ") || undefined;
+
+    const isEssentialAcrSatisfied = this.oidcAcr.isEssentialAcrSatisfied(
+      interaction,
+      activeUserSession,
+    );
 
     const canReuseActiveSession =
       isUserConnectedAlready &&
@@ -119,10 +126,6 @@ export class InteractionController {
       isSessionOpenedWithHintedLogin;
 
     const { name: spName } = await this.serviceProvider.getById(spId);
-    const { acrClaims } =
-      this.oidcAcr.getFilteredAcrParamsFromInteraction(interaction);
-    const spEssentialAcr =
-      acrClaims?.value || acrClaims?.values.join(" ") || null;
 
     if (!canReuseActiveSession) {
       userSession.clear();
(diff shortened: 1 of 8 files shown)


back/apps/core-fca-low/src/controllers/interaction.controller.ts
```diff
@@ -89,6 +89,8 @@ export class InteractionController {
       },
     } = interaction;
 
+    const { amr, idpAcr, isEmailVerifiedByPcf } = userSession.get();
```
rdubigny: suggestion: you should better not read the session before it is validated to avoid destructuration error or undefined variables.

back/libs/oidc-acr/src/oidc-acr.service.ts
```diff
@@ -166,4 +185,24 @@ export class OidcAcrService {
...
+    const { prompt } = interaction;
+    if (prompt.name !== "login") {
+      return false;
+    }
+
+    const containsEssentialAcrs =
+      prompt.reasons.includes("essential_acr") ||
+      prompt.reasons.includes("essential_acrs");
+
+    // if no acr is required, it is satisfied
```
rdubigny: todo: this comment makes no sense here. Move it L116.

back/libs/oidc-acr/src/oidc-acr.service.ts
```diff
@@ -107,16 +107,35 @@ export class OidcAcrService {
...
+    const { acrClaims } = this.getFilteredAcrParamsFromInteraction(interaction);
+    const spEssentialAcr =
+      acrClaims?.value || acrClaims?.values?.join(" ") || undefined;
+
+    const interactionAcr = this.getInteractionAcr({
+      idpAcr: userSession.idpAcr,
+      spEssentialAcr,
+      isEmailVerifiedByPcf: userSession.isEmailVerifiedByPcf,
+      amr: userSession.amr,
+    });
```
rdubigny, line 130: Thought: it is nice to reuse this sensitive function, but its usage still feels weird in this place.

`getInteractionAcr` should not return `undefined` if the ACR is not satisfied. If we want the current ACR, it should not depend on whether the SP asked for one or not.

Alternatively, this function should be called `isEssentialAcrSatisfied`, but you already used that name for the calling function, which raises more questions than it answers.

Maybe we should think this through a bit more. Note that in PCI, the function is called `areAcrsRequestedInPrompt`. Should we share some code between the two apps?

rdubigny (commented):
This does the job, except there are no end-to-end tests.

Can you add some?

This introduces some technical debt that we should address in another PR.

rdubigny (approved):

BenoitSerrano:
[Ne pas demander de nouvelle session ProCo (et donc de nouvel OTP mail) à chaque fois qu'un utilisateur se connecte a un service qui demande la MFA](https://trello.com/c/uPPQA0kG/1564-ne-pas-demander-de-nouvelle-session-proco-et-donc-de-nouvel-otp-mail-%C3%A0-chaque-fois-quun-utilisateur-se-connecte-a-un-service-qui)

PR #1451, opened 2026-07-31
🔧 make the local mongo watcher faster

douglasduteil:
**Problem**

Fast e2e tests can fail because the watcher wait for 1s before applying changes in memory in the core-fca-low. 

**Proposal**

Boost of **98,9%** from 1_000 to 11 to reduce edge cases


rdubigny (approved):
cool 👍🏻

PR #1452, opened 2026-07-31
🏗️ Simplify back Dockerfile multi-stage build

douglasduteil:
**Problem**

- `back/Dockerfile` indirected the base image through an unused `ARG NODE_VERSION`/`NODE_IMAGE` pair
- kept a separate `migrator` stage that duplicated `node_modules`
- rebuilt `node_modules` from scratch in `builder` instead of reusing the cached `dev_dependencies` layer
- threaded an unused `ARG INSTANCE` through `builder`/`production` despite only ever building `core-fca-low`
- copied the whole build context into `dev` via `COPY . ./` instead of naming what `yarn start:dev` actually needs
- the dedicated `migrator` image was still the only way `init-core` (`docker/compose/shared/compose.yaml`) could run `yarn run migrate up`

**Proposal**

- inline the node image tag
- base `builder` on `dev_dependencies` to reuse the cached `node_modules` layer
- hardcode `core-fca-low` as the only instance built from this Dockerfile
- fold `migrator` into `production` instead of dropping it: `migrate-mongo` is already a runtime dependency, so copying `package.json`, `migrate-mongo-config.js`, and `migrations/` lets `production` double as both app server and one-shot migration runner, matching CI's removal of the standalone `core-fca-low-migrator` build target
- point `init-core` at `target: production` accordingly
- replace `dev`'s `COPY . ./` with an explicit list of the files and directories `yarn start:dev core-fca-low` actually reads

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -94,7 +94,6 @@ jobs:
       matrix:
         instance:
           - core-fca-low
-          - core-fca-low-migrator
           - csmr-rie
           - mock-data-provider
           - mock-identity-provider
@@ -103,9 +102,6 @@ jobs:
         include:
           - instance: core-fca-low
             context: ./back
-          - instance: core-fca-low-migrator
-            context: ./back
-            target: migrator
           - instance: csmr-rie
             context: .
             file: ./csmr-rie/Dockerfile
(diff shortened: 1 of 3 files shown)


douglasduteil:
Extracted from #1439

BenoitSerrano (approved):
Wow, this change has been expected for a while!

PR #1453, opened 2026-07-31
🏗️ Run core migrations via pre_start hook, not init-core

douglasduteil:
**Problem**

`init-core` was a whole extra service — its own build, volumes,
env overrides, and a `depends_on: condition: service_completed_successfully`
gate — that existed only to run `yarn run migrate up` once before
`base-core` starts. Half of its config (`Mongoose_HOSTS`, TLS
settings) duplicated what `core.env` already sets on `base-core`
itself, since both talk to the same local `mongo`.

**Proposal**

Compose's `pre_start` hook runs inside the target container's own
lifecycle, before its main process starts, and inherits the
service's `environment`/`env_file` — so it doesn't need a
duplicate service or duplicate env wiring at all. Verified directly
(not just by reading docs) that a failing `pre_start` hook blocks
the container from ever reaching `Started`, matching the safety
`depends_on: service_completed_successfully` gave `init-core`, and
that a successful hook's output and env inheritance both work as
expected. Move the migration command onto `base-core` as
`pre_start: [command: yarn run migrate up]` and drop `init-core`
entirely — `core` now migrates itself before serving traffic.

### DIFF
--- a/docker/compose/shared/compose.yaml
+++ b/docker/compose/shared/compose.yaml
@@ -17,41 +17,9 @@ services:
       - data
     init: true
 
-  init-core:
-    build:
-      context: "../../../back"
-      target: production
-    working_dir: /var/www/app
-    user: ${CURRENT_UID}
-    depends_on:
-      mongo:
-        condition: service_healthy
-    volumes:
-      - "../../../back:/build/back:ro"
-      - "../../volumes/app:/opt/scripts"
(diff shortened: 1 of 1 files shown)


douglasduteil:
Extracted from https://github.com/proconnect-gouv/federation/pull/1439

BenoitSerrano (approved):
Very satisfying to read!

PR #1455, opened 2026-07-31
 🌱 Move seeding to the core image

douglasduteil:
**Problem**

Core dev-stack seeding was spread across mongo initdb and reset
scripts that diverged from the real application state, migrations ran
through a separate init-core step, and the admin CI stack reset the
mongo database but had no way to run the core seed because core is not
part of the admin stack.

**Proposal**

Centralize default data in `back/scripts/seed` (renamed for
consistency) and drive it from core's `yarn run seed`, run migrations
from a `pre_start` hook instead of init-core, and add a manual dev-stack
seed command. For the admin CI, run the seed through a one-off
`docker compose run --rm --no-deps core` container and pre-build the
core image in the stack step so the reset fits within the cypress exec
budget. Fold in the FQDN-to-attachedEmailDomain and
Fournisseur-de-données-to-resource-server renames, the email
verification/OTP rework, and dependency updates accumulated on the
branch.


back/scripts/seed/data/scopes.ts
```diff
@@ -0,0 +1,29 @@
+//
```
BenoitSerrano: The scopes are not needed anymore; they're hardcoded in the app and they've been deleted from the db

BenoitSerrano:
@gdarquie did a great job and renamed all the seeds in the [1481](https://github.com/proconnect-gouv/federation/pull/1481) PR (dp.js -> resource_servers.js, idp.js -> identity_providers.js, sp.js -> service_providers.js)
Can you rename them here as well for consistency?

douglasduteil:
Is this better @BenoitSerrano \cc @gdarquie ?

BenoitSerrano (approved):

PR #1458, opened 2026-08-03
🐛 fix: only resend verification email on explicit request

BenoitSerrano:
**Problem**
The verification email was automatically resent whenever the user refreshed the page after the cooldown period had expired, resulting in unnecessary emails.

**Proposal**
Only resend the verification email when the user explicitly clicks the "Request a new code" button.

**Girl scout**
Rename variable to verificationEmailCooldownBeforeResendInMs

### DIFF
--- a/back/apps/core-fca-low/src/controllers/email-verification.controller.ts
+++ b/back/apps/core-fca-low/src/controllers/email-verification.controller.ts
@@ -82,4 +82,28 @@ export class EmailVerificationController {
 
     return res.redirect(url);
   }
+
+  @Post(Routes.VERIFY_EMAIL_RESEND)
+  @Header("cache-control", "no-store")
+  @UseGuards(CsrfTokenGuard)
+  @UseFilters(EmailVerificationExceptionFilter)
+  async postResendVerifyEmail(
+    @Res() res: Response,
+    @UserSessionDecorator(AfterGetOidcCallbackSessionDto)
+    userSession: ISessionService<AfterGetOidcCallbackSessionDto>,
+  ): Promise<void> {
+    const {
+      spIdentity: { email },
+    } = userSession.get();
+
+    const { hasSentVerificationEmail } =
+      await this.emailVerification.sendEmailVerificationIfNeeded(email, {
+        requestSendEmail: true,
+      });
+
+    return this.emailVerification.renderVerificationEmailTemplate(res, {
+      email,
+      hasSentVerificationEmail,
+    });
+  }
 }
--- a/back/apps/core-fca-low/src/enums/routes.enum.ts
+++ b/back/apps/core-fca-low/src/enums/routes.enum.ts
@@ -10,6 +10,7 @@ export enum Routes {
   REDIRECT_TO_IDP = "/redirect-to-idp",
   IDENTITY_PROVIDER_SELECTION = "/identity-provider-selection",
   VERIFY_EMAIL = "/verify-email",
+  VERIFY_EMAIL_RESEND = "/verify-email/resend",
   OIDC_CALLBACK = "/oidc-callback",
   OIDC_LOGOUT_CALLBACK = "/client/logout-callback",
 }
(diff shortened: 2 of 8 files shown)


BenoitSerrano:
[Follow up : OTP mail](https://trello.com/c/cAxXObbR/1557-follow-up-otp-mail)

back/libs/email-verification/src/dto/email-verification-config.dto.ts
```diff
@@ -8,9 +8,9 @@ export class EmailVerificationConfig {
 
   @IsNumber()
   @IsPositive()
-  readonly tokenExpirationDurationInMs = 60 * 60 * 1000; // 1 hour in milliseconds
+  readonly tokenExpirationDurationInMs: number;
```
gdarquie, line 11: question: no more default value?

back/apps/core-fca-low/src/enums/routes.enum.ts
```diff
@@ -10,6 +10,7 @@ export enum Routes {
   REDIRECT_TO_IDP = "/redirect-to-idp",
   IDENTITY_PROVIDER_SELECTION = "/identity-provider-selection",
   VERIFY_EMAIL = "/verify-email",
+  RESEND_VERIFY_EMAIL = "/verify-email/resend",
```
gdarquie: For consistancy (we keep the resource "VERIFY_EMAIL" first and then add an action):

```suggestion
  VERIFY_EMAIL_RESEND = "/verify-email/resend",
```

Or EMAIL_VERIFICATION / EMAIL_VERIFICATION_RESEND ?

back/apps/core-fca-low/src/controllers/email-verification.controller.ts
```diff
@@ -82,4 +82,29 @@ export class EmailVerificationController {
 
     return res.redirect(url);
   }
+
+  @Post(Routes.RESEND_VERIFY_EMAIL)
+  @UsePipes(new ValidationPipe({ whitelist: true }))
```
gdarquie:
```suggestion
```

I think this pipe validation is not usefull here, ISessionService is an interface and not a DTO so the validation won't apply here.

I think this validation pipe is not useful here: ISessionService is an interface, not a DTO, so validation/whitelisting won’t effectively apply to userSession.

back/libs/email-verification/src/email-verification.service.ts
```diff
@@ -36,26 +36,30 @@ export class EmailVerificationService {
...
-        verificationEmailWaitingDurationBeforeResendInMs,
+        verificationEmailCooldownBeforeResendInMs,
     );
   }
 
-  async sendEmailVerificationIfNeeded(email: string) {
+  async sendEmailVerificationIfNeeded(
+    email: string,
+    option?: { requestSendEmail?: boolean },
+  ) {
```
gdarquie:
```suggestion
  ) : Promise<{ hasSentVerificationEmail: boolean }>{
```

back/libs/email-verification/src/email-verification.service.ts
```diff
@@ -36,26 +36,30 @@ export class EmailVerificationService {
...
       lastEmailVerificationTokenSentAt.getTime() +
-        verificationEmailWaitingDurationBeforeResendInMs,
+        verificationEmailCooldownBeforeResendInMs,
     );
   }
 
-  async sendEmailVerificationIfNeeded(email: string) {
+  async sendEmailVerificationIfNeeded(
+    email: string,
+    option?: { requestSendEmail?: boolean },
```
gdarquie:
```suggestion
    options?: { requestSendEmail?: boolean },
```

gdarquie (approved):

back/libs/email-verification/src/dto/email-verification-config.dto.ts
BenoitSerrano: Nope, I don't know why I put them in the first place, but it was the only file where we defined default values in the Dto config

BenoitSerrano:
[Ne pas demander de nouvelle session ProCo (et donc de nouvel OTP mail) à chaque fois qu'un utilisateur se connecte a un service qui demande la MFA](https://trello.com/c/uPPQA0kG/1564-ne-pas-demander-de-nouvelle-session-proco-et-donc-de-nouvel-otp-mail-%C3%A0-chaque-fois-quun-utilisateur-se-connecte-a-un-service-qui)

PR #1485, opened 2026-08-07
rename fqdns to attachedEmailDomains

BenoitSerrano:
**Problem**
When ProConnect Fédération sends an email OTP to the user for MFA, the Service Provider is not informed that an email verification was performed.

**Proposal**
Add `mail` to the `amr` claim returned to the Service Provider to indicate that the user's email address was verified.


BenoitSerrano:
[Renommer partout "fqdn" en "attachedEmailDomain"](https://trello.com/c/4GSXq0aS/1568-renommer-partout-fqdn-en-attachedemaildomain)

back/_doc/env-vars.md
```diff
@@ -10,65 +10,65 @@
...
+| App_DEFAULT_EMAIL_RENATER                        | string        |
+| App_DEFAULT_IDP_UID                              | string        |
+| App_DSFR_ASSETS_PATHS                            | json          |
+| App_FEATURE_DISPLAY_MAINTENANCE_NOTICE           | boolean       |
+| App_FEATURE_DISPLAY_TEST_ENV_WARNING             | boolean       |
+| App_HTTPS_SERVER_CERT                            | file          |
+| App_HTTPS_SERVER_KEY                             | file          |
+| App_MAINTENANCE_DATETIME                         | string        |
+| App_MAINTENANCE_DURATION                         | string        |
+| App_SP_AUTHORIZED_ATTACHED_EMAIL_DOMAINS_CONFIGS | json          |
```
gdarquie, line 32: We need to update env var in k8s -> k8s-proconnect/helm/fca-low/templates/deployment.yaml

back/apps/core-fca-low/src/interfaces/sp-authorized-fqdns-details.ts
```diff
@@ -1,5 +0,0 @@
-export interface SpAuthorizedFqdnsDetails {
```
gdarquie, line 1: question: it was not used anymore, you discovered it while cleaning, is that why you removed it?

gdarquie (approved):
Don't forget to rename `App_SP_AUTHORIZED_FQDNS_CONFIGS` to `App_SP_AUTHORIZED_ATTACHED_EMAIL_DOMAINS_CONFIGS`.

I understand the desire to rename `fqdn`, since it was too generic, but I wouldn't advocate for "email domain" either, since I think it's misleading.

As a reminder, for an email like "name@my.email.fr", "my.email.fr" is the FQDN of the email, and "email.fr" is the domain — we use the FQDN, not the domain.

I'm also not sure what an "attached" domain is supposed to mean here (attached to what?) (and "domaine email attaché" in French is a bit strange).

But I won't oppose it any further, so let's go for it.

back/apps/core-fca-low/src/interfaces/sp-authorized-fqdns-details.ts
BenoitSerrano: Exactly, sorry I did not specify it in the PR description!

BenoitSerrano:
I understand where you're coming from!
I think that in French that would be "domaine e-mail lié" rather than "attaché", and it refers to how an e-mail domain / fqdn is *bound* to an identity provider -> if you think "bound" / "linked" would be better than "attached", I'm not strongly opiniated on the matter

PR #1489, opened 2026-08-10
🔍 chore: add logs for duplicate OTP emails

BenoitSerrano:
**Problem**
Some users receive multiple one-time password emails, even though the algorithm should prevent this from happening within a 10-minute interval.

**Proposal**
Add logs around the OTP email generation and sending logic to investigate the cause of duplicate emails.

### DIFF
--- a/back/libs/email-verification/src/email-verification.service.ts
+++ b/back/libs/email-verification/src/email-verification.service.ts
@@ -62,6 +62,14 @@ export class EmailVerificationService {
       options?.requestSendEmail,
     );
 
+    this.logger.info({
+      code: "send-email-verification-if-needed",
+      lastEmailVerificationTokenSentAt:
+        lastEmailVerificationToken?.sentAt?.toISOString(),
+      requestSendEmail: options?.requestSendEmail,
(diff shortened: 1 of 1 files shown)


gdarquie (approved):

back/libs/email-verification/src/email-verification.service.ts
```diff
@@ -62,6 +62,14 @@ export class EmailVerificationService {
       options?.requestSendEmail,
     );
 
+    this.logger.info({
+      code: "send-email-verification-if-needed",
+      lastEmailVerificationTokenSentAt:
+        lastEmailVerificationToken?.sentAt?.toISOString(),
+      requestSendEmail: options?.requestSendEmail,
+      shouldSendEmail,
+    });
```
gdarquie, line 71: question: `lastEmailVerificationTokenSentAt`and `requestSendEmail` are custom properties and will be injected in pino by userContext?

BenoitSerrano: They are custom variables indeed but I don't know how pino handles it backstage

PR #1525, opened 2026-08-19
🧱 dev-stack: switch local TLS to mkcert nip.io domains

douglasduteil:
**Problem**

The local dev stack ships a committed self-signed CA and certificates for `*.docker.dev-franceconnect.fr`, a TLD that only works because the root CA is trusted manually on each machine. The stack also carries Cypress fixtures and a workflow for the retired integ01 environment.

**Proposal**

Move the local stack onto `*.proconnect.127.0.0.1.nip.io`, which resolves to 127.0.0.1 without hosts-file entries, and generate the TLS certificates on the fly with mkcert: the root CA lives in the local trust store, certs are gitignored, and `dks up` / a new `dks ssl` command refresh them when they expire. Drop the integ01 and k8s Cypress fixtures, their tag hooks, and the integration-tests workflow, and install mkcert in the CI stack jobs.


douglasduteil:
[ne plus utiliser le DNS record fc pour notre env de dev](https://trello.com/c/m8dXgFyx/1567-ne-plus-utiliser-le-dns-record-fc-pour-notre-env-de-dev)

gdarquie (commented):
Please divide this PR in many 🙏 The 100 files changed will kill me 🔪 
- one for the renamming of the routes
- one for removing all the tags and useless tests in cypress: https://trello.com/c/eT03fZ1I/1493-test-f%C3%A9d%C3%A9ration-supprimer-tous-les-tags-ignoredocker-k8s-ignoreinteg01

PR #1528, opened 2026-08-24
🐛 Recompute countdown from end date instead of decrementing on each tick

rebeccadumazert:
**Problem:**
setInterval decremented a local counter each tick instead of recalculating
from the end date, causing the countdown to drift over time.

**Fix:**
Recompute remaining seconds from the end date on every tick instead of
decrementing manually.


rebeccadumazert:
[Corriger le problème du shift du timer en front pour l'OTP mail](https://trello.com/c/kYwyG8J0/1611-corriger-le-probl%C3%A8me-du-shift-du-timer-en-front-pour-lotp-mail)

PR #1544, opened 2026-08-26
🔒️ Fix go-back link blocked by strict CSP script-src

rebeccadumazert:
**Problem:**
The go-back button used a `javascript:history.back()` href, which is blocked by the CSP's script-src directive since #1362 removed unsafe-inline.

**Fix:**
Replace the link with a `<button>` and move `history.back()` to an external script (go-back.js).


rebeccadumazert:
[Corriger le problème du bouton "retour" sur la page multi FI](https://trello.com/c/e9P6casa/1621-corriger-le-probl%C3%A8me-du-bouton-retour-sur-la-page-multi-fi)

gdarquie (approved):

PR #1546, opened 2026-08-27
💚 Fixing grist communication errors in CI

gdarquie:
**Context**

The CI fails consistently whenever upsert operations occur in the admin panel. When a new entity is created, a call is made to Grist expecting a list in response. However, Grist currently returns {"message": "document worker not present"} instead.

As a result, GristPublisherService attempts to iterate over this object, raising a previousGristRecords is not iterable error and failing the test suite. The API call to Grist itself succeeds, but the response body is an error payload rather than an iterable array.

**Proposed Solution**

This PR introduces a workaround to prevent hard failures—at least temporarily—when the Grist connection is established but no document worker is available.

- Updates the function to check the response structure and return false early, before any iteration is attempted.
- Relying on existing failure handling, returning false prevents the code from crashing and allows the application to gracefully display the standard "Grist sync failed" notification, without blocking the core entity creation event.
- Log the grist error message

### DIFF
--- a/admin/src/grist-publisher/grist-publisher.service.spec.ts
+++ b/admin/src/grist-publisher/grist-publisher.service.spec.ts
@@ -52,6 +52,47 @@ describe("GristPublisherService", () => {
   });
 
   describe("publishIdentityProviders", () => {
+    it("should return false when Grist records response does not contain a records array", async () => {
+      const mockGristConfig = {
+        ...baseMockGristConfig,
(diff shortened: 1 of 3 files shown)


gdarquie:
[Débloquer la CI Fédération](https://trello.com/c/MdpC5QI8/1627-d%C3%A9bloquer-la-ci-f%C3%A9d%C3%A9ration)

PR #1549, opened 2026-08-31
✨ Adding a fallback to get cachedOrganization when API are down

gdarquie:
**Context**

When API Entreprise is down, a large number of users in Internet-facing instances cannot connect because the API error is directly thrown, even though we have an Organization cache in MongoDB.

**Proposal**

When a cache for the requested organization exists, even if it is expired (as long as it does not exceed the max age accepted), catch the connection error from API Entreprise and return the cached value.

**About e2e tests**

I considered adding an e2e test for this fallback scenario, but opted against it due to complexity. quality/cypress currently only tests API reachability; simulating API Entreprise being down would require dedicated Docker Compose overrides or network mocks.

### DIFF
--- a/back/apps/core-fca-low/src/config/api-entreprise.ts
+++ b/back/apps/core-fca-low/src/config/api-entreprise.ts
@@ -10,6 +10,7 @@ const apiEntrepriseConfig: ApiEntrepriseConfig = {
   featureFetchOrganizationData: env.boolean("FEATURE_FETCH_ORGANIZATION_DATA"),
   organizationSiret: "13002526500013",
   cachedTTL: 86400000, // 24h
+  cacheTTLWhenApiEntrepriseIsDown: 7776000000, // 90 days
 };
 
 export default apiEntrepriseConfig;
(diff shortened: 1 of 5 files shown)


gdarquie:
[Utiliser la valeur en cache des organisations si possible lorsque l'API SIRENE est down](https://trello.com/c/U8O0FtAg/1612-utiliser-la-valeur-en-cache-des-organisations-si-possible-lorsque-lapi-sirene-est-down)

back/apps/core-fca-low/src/config/api-entreprise.ts
```diff
@@ -10,6 +10,7 @@ const apiEntrepriseConfig: ApiEntrepriseConfig = {
   featureFetchOrganizationData: env.boolean("FEATURE_FETCH_ORGANIZATION_DATA"),
   organizationSiret: "13002526500013",
   cachedTTL: 86400000, // 24h
+  maxAgeForCache: 7776000000, // 90 days
```
BenoitSerrano: wrong name

BenoitSerrano (approved):
Approved provided you change this name!

PR #1550, opened 2026-09-01
♻️ Add rewrites and clarifications

rdubigny:
**Problem**

`spAmr` has a different lifecycle than `interactionAcr`, despite serving a very similar purpose.

`deleteEmailVerificationToken` exposes internal logic that should be handled by the email verification service.

`idp_acr` in `identityForSp` looks like it can be used like the `idpAcr` field in the user session, but it serves another, deprecated purpose.

**Proposal**

Rename `spAmr` to `interactionAmr`. Set its value during interaction verification and create a dedicated function to assign it.

Delete the consumed token at the end of the email verification flow.

Clarify the usage of the `idp_acr` attribute and mark it as deprecated to prevent future use.


BenoitSerrano (approved):
Very elegant, thanks!

PR #1577, opened 2026-09-02
🐛 Handle API Entreprise downtime correctly

gdarquie:
**Context**

When API Entreprise is down, `ApiEntrepriseConnectionException` is supposed to be thrown and the special `CachedOrganization` cache should be used, but that is not currently the case.

**Proposal**

In this PR, we catch an additional error to ensure the expected behavior.

**Manual tests**

I tested locally:
- When I disable the mock for ApiEntreprise_SHOULD_MOCK_API, I clearly see the dedicated error Y500028.
- When a cache exists and is less than 24 hours old, I can still log in without error when the API Entreprise token is invalid.
- When a cache has existed for more than 90 days, the Y500028 error is thrown.
- When a cache has existed for more than 24 hours but less than 90 days, I can still log in without error.
- When there is no cache, the Y500028 error is thrown.


gdarquie:
[Follow up API Entreprise / SIRENE](https://trello.com/c/dcdnxv3S/1644-follow-up-api-entreprise-sirene)

BenoitSerrano (approved):

PR #1590, opened 2026-09-03
🔥 remove unused packages

rdubigny:


BenoitSerrano (approved):
Aaah now we can breathe

PR #1593, opened 2026-09-08
✨ feat: preselect the last used identity provider

BenoitSerrano:
**Problem**
When a user's email domain is associated with multiple Identity Providers, they have to select their Identity Provider again at each login.

**Proposal**
Preselect the last Identity Provider used by the user.

### DIFF
--- a/back/apps/core-fca-low/src/public/js/multifi-preselection.js
+++ b/back/apps/core-fca-low/src/public/js/multifi-preselection.js
@@ -1,0 +1,33 @@
+const LOCAL_STORAGE_LAST_FI_SELECTED = "proConnectLastFISelected";
+
+function init() {
+  const localStorageLastFISelected = localStorage.getItem(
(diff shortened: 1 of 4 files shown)


BenoitSerrano:
[Présélectionner le choix du FI sur la mire multi-FI](https://trello.com/c/YgXwCbHI/1513-pr%C3%A9s%C3%A9lectionner-le-choix-du-fi-sur-la-mire-multi-fi)

back/apps/core-fca-low/src/public/js/multifi-preselection.js
```diff
@@ -0,0 +1,33 @@
+const LOCAL_STORAGE_LAST_FI_SELECTED = "proConnectLastFISelected";
+
+function init() {
```
rdubigny: question: is there a reason that explain why you are not waiting for the event "DOMContentLoaded" to be fired for this particular script?

rdubigny (approved):

PR #1623, opened 2026-09-11
🐛 fix: seed admin fixtures locally

BenoitSerrano:
**Problem**
Admin fixtures are no longer seeded automatically in the local environment.

**Proposal**
Add the fixture seeding command to `up.sh`.


douglasduteil (approved):

docker/bash/commands/up.sh
```diff
@@ -30,6 +30,8 @@ _up() {
       ${DOCKER_COMPOSE} exec ${NO_TTY} "admin" yarn typeorm schema:drop
       ${DOCKER_COMPOSE} exec ${NO_TTY} "admin" yarn migrations:run
       ${DOCKER_COMPOSE} exec ${NO_TTY} "admin" yarn fixtures:load
+      ${DOCKER_COMPOSE} run --rm --no-deps ${NO_TTY} core yarn run seed
```
rdubigny, line 33: @BenoitSerrano this instruction should not be under the case `pg-admin` nor the echo `Fixtures for admin app`.

PR #1625, opened 2026-09-14
🚨 add artifact-metadata permission to docker build workflow

douglasduteil:
**Problem**
build-docker.yml build job was missing the artifact-metadata:write permission, causing attest-build-provenance to fail creating the attestation storage record for every image/arch.

**Proposal**
Add artifact-metadata: write to the job's permissions block, alongside the existing attestations: write and id-token: write, so the storage record can persist.


douglasduteil:
example https://github.com/proconnect-gouv/federation/actions/runs/34820837459
image

.github/workflows/build-docker.yml
```diff
@@ -20,6 +20,7 @@ jobs:
       contents: read
       attestations: write
       id-token: write
+      artifact-metadata: write
```
rdubigny, line 23: suggestion: we should give more grannular permissions at step-level.

rdubigny (approved):

PR #1627, opened 2026-09-14
✨ feat: add e2e test harness for core-fca-low (TestingBench)

douglasduteil:
**Problem**
core-fca-low has no way to boot the real AppModule in tests against a real Mongo (needed for $changeStream support) instead of mocked services. An earlier prototype #677  went stale (979 commits behind main, one dependency it relied on no longer exists).

**Proposal**
Port the harness piece only, first of three planned stacked branches: TestingBench boots AppModule with mongodb-memory-server + ioredis-mock and exposes Symbol.asyncDispose, TEST_CONFIG rebuilt against the current CoreFcaConfig shape, smoke spec proving boot + clean teardown. Mongo fixtures and the interaction-controller e2e spec are follow-up branches.

PR #1645, opened 2026-09-16
✨ port oidc-client.controller e2e coverage against TestingBench

douglasduteil:
**Problem**
oidc-client.controller.ts (identity-provider-selection, redirect-to-idp, oidc-callback, client/logout-callback) only had unit-level coverage with every dependency mocked, so real session mutations, real IdP/account lookups, and real redirect construction were never exercised end-to-end, unlike interaction.controller and email-verification.controller.

**Proposal**
Added an e2e spec driving TestingBench against the real Mongo-backed CoreFcaService, AccountFcaService, and IdentitySanitizer, mocking only the external IdP boundary (OidcClientService), matching this repo's existing e2e pattern; retired the 6 of 11 unit cases whose branch and asserted value are now proven for real by an e2e case, keeping the 5 that cover scenarios (Entra claim shapes, an omitted rememberMe field) no e2e case currently exercises.


douglasduteil:
Superseded: re-cut from `douglasduteil/test-bench-epic-sweep`'s final state into per-route oidc-client e2e PRs (train tickets T8/T9). Never reviewed; closing to avoid reviewing stale diffs. #1627 (harness, approved) stays.

PR #1647, opened 2026-09-18
🔒 Remove PAR URL from discovery metadata

rdubigny:
**Problem**

PAR has been enabled by default since `oidc-provider` v8.

See
`https://github.com/panva/node-oidc-provider/blob/main/CHANGELOG.md#800-2022-12-03`.

However, we did not open the corresponding route in `oidc-provider.controller`, so the endpoint keeps returning a 404.

**Proposal**

Disable this endpoint in `oidc-provider` to reflect the actual state of the application.

We considered opening this feature, but since no one has requested it, we would rather reduce the number of exposed endpoints for security reasons.


rdubigny:
[Enquêter sur la disparition du endpoint pushed_authorization](https://trello.com/c/oJwcQAo0/1692-enqu%C3%AAter-sur-la-disparition-du-endpoint-pushedauthorization)

BenoitSerrano (approved):

